Issue #19, opened 2018-06-20
Interface: add functionality for dealing with multiple syntactic inputs (c-sets)


jbellik:
The current interface can only make a VT with a single candidate set and syntactic input, but most analyses require several different syntactic inputs, necessitating tedious copy-pasting for the final compilation of c-sets.

PR #54, opened 2019-07-11
Multiple inputs in gui

ozanblk:
#19


jbellik:
All the functionality seems good, although visually and UX-wise, it's inconvenient to have the trees arranged vertically instead of horizontally. Scrolling between lower trees and the "Add mother" etc buttons gets very inconvenient. This could be ameliorated by putting new trees at the top of the list of trees instead of the bottom, but would ideally be remedied by lining them up horizontally.

jbellik:
All the functionality seems good, although visually and UX-wise, it's inconvenient to have the trees arranged vertically instead of horizontally. Scrolling between lower trees and the "Add mother" etc buttons gets very inconvenient. This could be ameliorated by putting new trees at the top of the list of trees instead of the bottom, but would ideally be remedied by lining them up horizontally.

PR #68, opened 2019-07-22
Bin Rightmost

nnnguyenhoa:


constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+      }       
+    }
+    //check other nodes in ptree
+    for(var i = 0; i < ptree.children.length; i++) {
+      vcount += binMaxRightmostBranches(s, ptree.children[i], cat);
+    }       
+  }
+  return vcount;
+};
+var parent_ptree;
```
jbellik: Define this variable inside the function so the function is self-contained (better code hygiene).

constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+function binMaxRightmostLeaves(s, ptree, cat) {
+  //first find the first level of with cat --> recursive look into the children
+  var vcount = 0;
+  if(ptree.children && ptree.children.length) {
+  if(ptree.cat === cat) {
```
jbellik: indentation issue here

constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+      if(ptree.children.length > 2) {
+      	if(parent_ptree && ptree === parent_ptree.children[parent_ptree.children.length-1]) {
+        	vcount++;
+      	}
+      }
+    }
+  }
+  for(var i = 0; i < ptree.children.length; i++) {
+  	parent_ptree = ptree;
+    vcount += binMaxRightmostLeaves(s, ptree.children[i], cat);
```
jbellik: indentation problem

constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+};
+var parent_ptree;
+function binMaxRightmostLeaves(s, ptree, cat) {
+  //first find the first level of with cat --> recursive look into the children
+  var vcount = 0;
+  if(ptree.children && ptree.children.length) {
+  if(ptree.cat === cat) {
+    var recur_flag = 0;
+    for(var i = 0; i < ptree.children.length; i++) {
+      if(ptree.children[i].cat === cat) {
```
jbellik: Here vcount is being incremented without checking whether there are more than 2 nodes of cat - 1 dominated. For example, let toomanyvtree = {((a))((b))((c))}, where {} stands for iota and () stands for phi, and a, b, c are ws. Then binMaxRightmostLeaves({}, toomanyvtree, "phi") should be 0. But the current implementation incorrectly returns 3 for this case. 
This check has nothing to do with whether there is actually a violation of the constraint.

constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+        vcount++;
+	recur_flag = 1;
+	break;
+      }
+    }
+    if(recur_flag === 0) {
+      if(ptree.children.length > 2) {
+      	if(parent_ptree && ptree === parent_ptree.children[parent_ptree.children.length-1]) {
```
jbellik: There is no guarantee parent_ptree is correctly defined. One issue is that for the root node, parent_ptree will either be undefined or incorrectly defined (left over from a previous call to this function). The root node should automatically be considered rightmost. For example, according to the constraint definition, binMaxRightmostLeaves({}, {(a)(b)(c)}, "i") should return 1, since the iota dominates 3 phis. But instead it returns 0.

constraints/binRightmost.js
```diff
@@ -0,0 +1,60 @@
...
+        	vcount++;
+      	}
+      }
+    }
+  }
+  for(var i = 0; i < ptree.children.length; i++) {
+  	parent_ptree = ptree;
+    vcount += binMaxRightmostLeaves(s, ptree.children[i], cat);
+  }
+  //once found save this level to look at other children for same cat
```
jbellik: Is this comment up-to-date? I don't think it describes what is actually happening.

test/binmaxRight.js
```diff
@@ -0,0 +1,79 @@
+function binaryRightmost(stree, ptree, cat)
```
jbellik: Is this file even being used? I think it is leftover from the fall or winter. It's not syntactically valid.

trees/trees.js
```diff
@@ -104,6 +104,33 @@ var prostree3 = {
 	]
 };
 
+var ex_violation_tree = {
```
jbellik: Is this the file you meant to add this tree to?

PR #84, opened 2019-08-13
Equal boundaries

maxTarlov:
EqualBoundaries constraints as proposed by Nick. equalStrengthBase is called by the constraints and equalStrengthHelper is called by equalStrengthBase. equalStrengthHelper recursively goes through each node on the tree and calls the nested function recursiveStrengthHelper, which recursively goes through the left or right branches of a subtree, as specified in the argument structure of the outer function. recursiveStrenghtHelper records the number of nodes it encounters that match the category specified in the argument structure and, when it reaches a terminal, stores that number in the terminal with an object property called "edges", if that terminal does not already have a value for edges. Edges is always cleared out once in equalStrengthBase so that edges is assigned once and only once during the protocol. equalStrengthBase returns an array containing the terminals of the syntactic tree and the terminals of the prosodic tree. The constraints then compare the number of edges in each terminal and assign the appropriate number of violations.


build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+  }
+}
+
+/* A function to return the paired category as defined in categoryPairings.
+ * categoryPairings only returns prosodic categories given a syntactic category.
+ * reversableCatPairings also returns a syntactic category given a prosodic
+ * category.
+ * Maybe this should move to prosodicHierarchy.js?
+*/
+function reversableCatPairings(cat){
```
jbellik: reversible, not reversable.

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+      equalStrengthHelper(tree.children[i], cat, edgeName);
+    }
+  }
+}
+
+/* A function to return the paired category as defined in categoryPairings.
+ * categoryPairings only returns prosodic categories given a syntactic category.
+ * reversableCatPairings also returns a syntactic category given a prosodic
+ * category.
+ * Maybe this should move to prosodicHierarchy.js?
```
jbellik: Yes, good idea.

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+function reversableCatPairings(cat){
+  if (categoryPairings[cat]){
+    return categoryPairings[cat]; //just the same as calling categoryPairings
+  }
+  else {
+    //get the property names of categoryPairings
+    var props = Object.getOwnPropertyNames(categoryPairings);
+    var propFound = false; //true when the category is paired
+    for (var i = 0; i < props.length; i ++){
+      if (categoryPairings[props[i]] == cat){
```
jbellik: === for strict equality

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+        propFound = true;
+        if (props[i] == "clause"){
+          // rn categoryPairings has a property "clause" which maps to i
+          return "cp"; // "cp" also maps to i, I think we want "cp"
+        }
+        //props[i] is the property that maps to cat
+        return props[i];
+      }
+    }
+    // if no matching category is found, return a costom error.
```
jbellik: custom not costom

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+    if (!propFound){
+      var err = new Error("" + cat + " is not a category defined in categoryPairings (see main/prosodicHierarchy.js)");
```
jbellik: Throw this error instead of returning it:
throw(new Error("blah blah"));

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+  equalStrengthHelper(ptree, pcat, edgeName);//call recursive helper for prosody
+
+  //return an array of the terminals, also arrays, now with the property edges
+  return [getLeaves(stree), getLeaves(ptree)];
+}
+
+/* equalStrengthBase is run once per call so that sTerminals and pTerminals can
+ * be cleared out once. equalStrengthHelper runs recursively, once for each node
+ * on the tree. Also, the same process needs to be run on ptree and stree, so
+ * the helper function minimizes code repitition by taking only one tree and
```
jbellik: repetition

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+ * being called for both ptree and stree.
+ */
```
jbellik: This comment provides good information, but please expand it to say what the function is supposed to do. For example, "Takes a node [of what sort?] and returns [what information?]"

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+    if (edgeName == "left"){
+      edgeIndex = 0;
+    }
+    //right = [length -1]
+    if (edgeName == "right" && partree.children && partree.children.length){
+      edgeIndex = (partree.children.length - 1);
+    }
+    //increment boundaries when partree is of the correct category
+    if (partree.cat === category){
+      boundaries ++;
```
jbellik: Conventionally, there is no space between the variable name and the ++.

build/constraints.js
```diff
@@ -849,7 +928,189 @@ function splitNMCmin(sTree,pTree)
...
+    }
+    //if partree is terminal and partree.edges is not already defined
+    //edges must not already be defined because function is called recursively
```
jbellik: I'm not sure I understand this comment... Should like 992 say "then edges must be defined" or "needs to be defined"?

constraints/equalStrength.js
```diff
@@ -0,0 +1,182 @@
+/* Equal Strength Boundaries Constraints
```
jbellik: Max, I apologize for all the comments that should be on this file but which I put on the build/constraints.js file instead...

constraints/equalStrength.js
```diff
@@ -0,0 +1,182 @@
...
+    }
+    // if no matching category is found, return a costom error.
+    if (!propFound){
+      var err = new Error("" + cat + " is not a category defined in categoryPairings (see main/prosodicHierarchy.js)");
+      return err;
+    }
+  }
+}
+
+/* Equal Strength Right Syntax --> Prosody:
```
jbellik: These all look great. Very readable. Good modularity.

constraints/equalStrength.js
```diff
@@ -0,0 +1,182 @@
+/* Equal Strength Boundaries Constraints
+ * as proposed by Nick Kalivoda, August 2019:
+ * Assign violations if a phonological terminal is at the left/right edge of a
+ * different number of syntactic constituents as phonological constituents.
+ * Specific functions defined below. Base function counts number of left/right
+ * edges a terminal falls on.
+ */
+
+ //this code isn't robust for terminal order-shifting, does it need to be?
```
jbellik: I think it's okay that it's not robust to reordering of the terminals. Since we don't have a good way to refer to edges directly (they are not objects), making these functions robust to reordering would be a huge pain.

test/equalStrengthLeftTests.html
```diff
@@ -0,0 +1,52 @@
...
+		<link rel="stylesheet" type="text/css" href="../spot.css">
+		<script src="../build/spot.js"></script>
+		<script src="../build/constraints.js"></script>
+		<script src="../candidategenerator.js"></script>
+	</head>
+	<body>
+		<pre id="results-container"></pre>
+
+		<script>
+			var sTree ={
```
jbellik: Is this the only stree you tested with? I just tested it with an stree with more stacked up edges: [[[[a]b]c]d] vs. {((a)b) c d} and that seemed to work, but it would be good to test with some additional, methodically constructed stree/ptree combinations as well.

Definitely check whether it's working properly with different categories, too, because my preliminary test with {{(a) b} c d} + [[[[a]b]c]d] did not give me any errors.

test/equalStrengthRightTests.html
```diff
@@ -0,0 +1,52 @@
+<!-- Max Tarlov 8/13/19 testing abstract match constraints (Kalivoda) -->
+<html>
```
jbellik: I think it would be easier to compare the effects of the Left / Right constraints if you put them in the same tableau / test file.

build/constraints.js
maxTarlov: changed in latest commit

build/constraints.js
maxTarlov: moved as of latest commit

build/constraints.js
maxTarlov: changed as of latest commit

build/constraints.js
maxTarlov: changes as of latest commit

build/constraints.js
maxTarlov: changed in latest commit

build/constraints.js
maxTarlov: changed in latest commit

build/constraints.js
maxTarlov: comment expanded as of latest commit

build/constraints.js
maxTarlov: oh, well I tried to change some of them but I have been putting spaces in everywhere so I didn't change all of them. Thanks for the info, though

build/constraints.js
maxTarlov: I tried to clarify this in the latest commit. Basically it is the same as with markMinMax, where it does not write to the property unless it is undefined to avoid problems with recursive calls.

constraints/equalStrength.js
maxTarlov: I actually realized that this code might be robust for order-changing analyses, see the email chain "Abstract Match Constraints Update." I removed the comment to avoid further confusion, though.

test/equalStrengthLeftTests.html
maxTarlov: So I tested with a lot of different trees, but mostly in console so they aren't represented here. It looks like what happened was that you used a prosodic category when the function required a syntactic category, which doesn't break the code, it just means that the base functions are counting the number of iotas syntactic terminals are falling on and the number of CPs prosodic categories are falling on.
Since this is a mistake that I also made in testing, I added a little code bloc to convert prosodic categories to syntactic categories so that either can be used in equalStrengthLeft and equalStrengthRight.

test/equalStrengthRightTests.html
maxTarlov: merged two files in latest commit

maxTarlov:
As I said in my comment above, I think you just called equalStrengthLeft with the wrong type of category and that I added some code so that this won't be an issue anymore.
I was wondering if there is a way to put multiple tableau in one test file so that I can test with more inputs without adding more files or deleting previous tests? My understanding is that writeTableau looks for an HTML element called "results container" and puts the tableau there, so if I ran writeTableau again, it would just overwrite the first one, right? Maybe I'm missing something.
Hope the updates helped! As always, let me know if you have any questions.

jbellik:
Yes, definitely. Take a look at the file SPOT_custom_analysis_template.HTML
or any of the files in the folder analysis_HTML_files for examples. I can
walk you through one on Monday after we talk to Nick if you like; if you
want to do that then let's meet on campus after all. Have a look and let me
know.

It's very possible I used the wrong category/constraint combo as you said.
Great idea to make the code compensate for that error. Maybe instead of
actually defining separate SP and PS constraints for these, it would make
sense to have the direction determined by the category argument? Let's talk
about that on Monday too.

Thanks, Max!!

PR #86, opened 2019-08-21
implements strongStartClitic from Bennett et al. 2016

jbellik:
This implements the version of StrongStart from Bennett et al. 2016 under the name strongStartClitic. Refer to strongStartTest.html to check the violation counts -- open it in a browser and look at the rightmost columns.


rbibbs:
The violations for 1-5 look good as far as I can tell. However, I'm a little unsure about what the "w"s represent such that they incur violations for SS-i in candidates 6 & 7.

PR #87, opened 2019-08-22
Clitic randomizer

jbellik:
This code calls on GEN with the word order changed by moving the clitic to all possible positions in the prosodic tree. It is intended for use in analyzing pronoun placement in Chamorro.


build/spot.js
```diff
@@ -2725,6 +2725,86 @@ function addPhiWrapped(candidates, options){
...
+		if(cliticTagIndex > 0){
+			var wordListParts = wordList.split("-clitic");
+			wordList = wordListParts[0]+wordListParts[1];
+		}
+		wordList = wordList.split(' ');
+	}
+	//Find the clitic to move around
+	var cliticIndex = wordList.indexOf(clitic);
+	if(cliticIndex < 0)
+		throw new Error("The provided clitic"+clitic+" was not found in the word list");
```
ozanblk: Would be good to have a space between between the string "clitic" and the clitic.

build/spot.js
```diff
@@ -2725,6 +2725,86 @@ function addPhiWrapped(candidates, options){
...
+	//Slice the clitic out
+	var beforeClitic = wordList.slice(0,cliticIndex);
+	var afterClitic = wordList.slice(cliticIndex+1, wordList.length);
+	var cliticlessWords = beforeClitic.concat(afterClitic);
+
+	var orders = new Array(wordList.length);
+	for(var i = 0; i < wordList.length; i++){
+		beforeClitic = cliticlessWords.slice(0,i);
+		afterClitic = cliticlessWords.slice(i, cliticlessWords.length);
+		orders[i] = beforeClitic.concat([clitic+"-clitic"].concat(afterClitic));
```
ozanblk: You can can pass multiple arrays to concat

build/spot.js
```diff
@@ -2725,6 +2725,86 @@ function addPhiWrapped(candidates, options){
...
+	var cliticIndex = wordList.indexOf(clitic);
+	if(cliticIndex < 0)
+		throw new Error("The provided clitic"+clitic+" was not found in the word list");
+	//Slice the clitic out
+	var beforeClitic = wordList.slice(0,cliticIndex);
+	var afterClitic = wordList.slice(cliticIndex+1, wordList.length);
+	var cliticlessWords = beforeClitic.concat(afterClitic);
+
+	var orders = new Array(wordList.length);
+	for(var i = 0; i < wordList.length; i++){
```
ozanblk: Looks like the case where the clitic is the last item is not generated? Is this intentional?

candidategenerator.js
```diff
@@ -228,3 +228,80 @@ function addPhiWrapped(candidates, options){
...
+	//Otherwise, get the clitic from words
+	else
+	{
+		// Make sure words is an array
+		if(typeof words === "string"){
+			words = words.split(' ');
+		}
+		var x = words.find(containsClitic);
+		clitic = x.split('-clitic')[0];
+		words[words.indexOf(x)] = clitic;
```
ozanblk: there is a findIndex method, as well (could use it instead of find and indexOf)

candidategenerator.js
```diff
@@ -228,3 +228,80 @@ function addPhiWrapped(candidates, options){
...
+	}
+
+	//Make sure words is defined before using it to generate word orders
+	if(words.length<leaves.length){
```
ozanblk: if(!words || words.length < leaves.length) {

build/spot.js
jbellik: fixed

build/spot.js
jbellik: It is generated because wordList.length = cliticlessWords.length +1. But I changed this for readability.

PR #112, opened 2019-09-18
Built in Irish Analysis

maxTarlov:
Unless I am forgetting something, I think it all works. Some notes:

- We could stand to tinker with the layout around the built-in analysis, it looks a little clunky right now.
- I am thinking that maybe the built-in analysis stuff will look less out-of-place if it is in a fieldset, but that might also just make new users less likely to see it.
- I think there is an issue with the Irish tones, I'll email Jenny.


jbellik:
This built-in system looks good. It seems there is a problem with the Irish tone annotation still, which seems to be an issue specific to the interface, but that is unrelated to the code here.

PR #122, opened 2019-09-27
Recursive categories gen

maxTarlov:
User can now specify the root category, the recursive (intermediate) category and the terminal category. Also Jenny added a GEN function to create only unary nodes.


candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+		}
+		else{
+			options.rootCategory = "i";
+		}
+	}
+
+	//set default terminal category based on options passed
+	if(!options.terminalCategory){ //terminalCategory is unspecified
+		//terminalCategory unspecified, recursiveCategory specified as "i"
+		if(options.recursiveCategory === "i"){
```
jbellik: Can we abstract this? Instead of using 3 conditionals, use pcat.nextLower(options.recursiveCategory)
Recursive category is guaranteed to be specified at this point since you set it in line 44.

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+		else if(options.recursiveCategory === "w"){
+			options.terminalCategory = "Ft";
+		}
+		//neither terminalCategory nor recursiveCategory is specified, default to "w"
+		else{
+			options.terminalCategory = "w";
+		}
+	}
+
+	/* the prosodic hierarchy should include the categories specified in
```
jbellik: Expand this comment to spell out that if the user enters a category that isn't already in the prosodic hierarchy, the prosodic hierarchy will be amended to add it.

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+	 * options.rootCategory, options.recursiveCategory and options.terminalCategory
+	 */
+	if(options.rootCategory || options.recursiveCategory || options.terminalCategory){
+		//the specified root category should probably be highest
+		if(pCat.indexOf(options.rootCategory)<0)
```
jbellik: Let's also give a warning saying "The specified root category "+rootCategory+" is not in SPOT's pre-defined prosodic hierarchy (see pCat in prosodicHierarch.js). It is being added to your local version of the prosodic hierarchy."

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+			pCat.unshift(options.rootCategory);
+		if(pCat.indexOf(options.recursiveCategory)<0 && options.rootCategory !== options.recursiveCategory){
+			//the specified recursive category should be after iota
```
jbellik: Again, there should be a warning that the requested recursive category wasn't in pCat.
I'm not really so sure that GEN should be able to modify pCat like this anyway...

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+			try{
+				var afterI = pCat.indexOf("i") + 1;
+				pCat.splice(afterI, 0, options.recursiveCategory);
+			}
+			catch(err){
+				console.error("make sure pCat contains your recursive category.");
```
jbellik: This error message should spell out that a) your recursive category wasn't in pCat, b) we tried to add your recursive category to pCat but it failed for some reason.

candidategenerator.js
jbellik: I think that if the user enters unrecognized categories, we should actually just throw an error and not give any output, because it's impossible to know where in the prosodic hierarchy they should really be added. I see that you've made some logical choices about where in the hierarchy they could be added but they still seem somewhat arbitrary to  me, and will certainly yield a very opaque prosodic hierarchy! Let's pull out this code and stash it somewhere. It could be good to add functionality to let users create custom prosodic hierarchies and *then* be able to use them in their GEN.

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+		//the specified terminal category should be right above "syll"
+		if(pCat.indexOf(options.terminalCategory)<0 && options.terminalCategory !== options.recursiveCategory && options.rootCategory !== options.terminalCategory)
+			pCat.splice(-1, 0, options.terminalCategory);
+	}
+
+	if(options.rootCategory === options.recursiveCategory && options.obeysNonrecursivity){
+		alert("Warning:\nYou have instructed GEN to produce non-recursive trees and to produce trees where the root node and\nintermediate nodes are of the same category. Some of the trees GEN produces will be recursive.")
+	}
+
+	if(options.rootCategory === options.recursiveCategory && options.obeysNonrecursivity){
```
jbellik: Should this be (options.rootCategory === options.terminalCategory)?

jbellik: Also, in this case, *all* of the trees will be recursive.

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+		alert("Warning:\nYou have instructed GEN to produce non-recursive trees and to produce trees where the root node and\nterminal nodes are of the same category. Some of the trees GEN produces will be recursive.")
+	}
+
+	if(options.recursiveCategory === options.terminalCategory && options.obeysNonrecursivity){
+		alert("Warning:\nYou have instructed GEN to produce non-recursive trees and to produce trees where the intermediate\nnodes and the terminal nodes are of the same category. You will only get one bracketing.")
```
jbellik: Lol. This is a good warning.

candidategenerator.js
```diff
@@ -61,23 +129,29 @@ window.GEN = function(sTree, words, options){
...
+	 * only one child, which will be of the same category, ie. {i {i (...) (...)}}
+	 */
+	var rootlessCand = gen(leaves, recursiveOptions)
+	if(options.rootCategory !== options.recursiveCategory)
+	 rootlessCand = addRecCatWrapped(gen(leaves, recursiveOptions), options);
 
 	var candidates = [];
 	for(var i=0; i<rootlessCand.length; i++){
-		var iota = iotafy(rootlessCand[i], options);
+		var iota = wrapInRootCat(rootlessCand[i], options);
```
jbellik: This variable name is still misleading, I'm afraid. Can you change it to proot or root or something else category-neutral?

candidategenerator.js
```diff
@@ -98,10 +172,10 @@ window.GEN = function(sTree, words, options){
 	return candidates;
 }
 
-function iotaIsHeaded(iota) {
+function rootIsHeaded(iota, recCat) {
 	var children = iota.children || [];
 	for (var i = 0; i < children.length; i++)
-		if (children[i].cat === 'phi')
+		if (children[i].cat === recCat)
```
jbellik: If someone enters 'i' for the root category and 'w' for the recursive category, then this will yield the wrong results, since we'd be looking for a 'w' to be the head of the 'i'. I think we should be checking for a child of category pCat.nextLower(iota.cat) [except that I want iota to get renamed].

candidategenerator.js
```diff
@@ -28,14 +28,82 @@ var wNum = 0;
...
+	}
+
+	if(options.rootCategory === options.recursiveCategory && options.obeysNonrecursivity){
+		alert("Warning:\nYou have instructed GEN to produce non-recursive trees and to produce trees where the root node and\nterminal nodes are of the same category. Some of the trees GEN produces will be recursive.")
+	}
+
+	if(options.recursiveCategory === options.terminalCategory && options.obeysNonrecursivity){
+		alert("Warning:\nYou have instructed GEN to produce non-recursive trees and to produce trees where the intermediate\nnodes and the terminal nodes are of the same category. You will only get one bracketing.")
+	}
+
```
jbellik: We need a check and warning somewhere in here if recCat isn't the next category down the prosodic hierarchy from rootCat. E.g., if someone asks for trees rooted in iota with w as the recursive category, there aren't going to be any phis in the results. That is probably not what they have in mind. Every single tree will be violating Exhaustivity in a major way.

candidategenerator.js
```diff
@@ -61,23 +129,29 @@ window.GEN = function(sTree, words, options){
...
 	phiNum = wNum = 0;
 	for(var i=0; i<words.length; i++){
-		leaves.push(omegafy(words[i]));
+		leaves.push(wrapInLeafCat(words[i], options.terminalCategory));
 	}
 
 	var recursiveOptions = {};
 	for (var k in options) {
-		if (options.hasOwnProperty(k) && k !== 'requirePhiStem')
+		if (options.hasOwnProperty(k) && k !== 'requireRecWrapper')
```
jbellik: We should actually talk about whether this option is useful now that it's possible to specify the desired root category.

candidategenerator.js
jbellik: We also need a check for situations that can create layering violations. E.g., the user specifies {rootCategory:'i', terminalCategory: 'i'}, but doesn't specify recursiveCategory. Then by default, recursiveCategory is set to 'phi' and you end up with trees where the root and leaves are iotas but the intermediate nodes are phis. This merits a warning or error.

Basically at the end of all the category setting, there needs to be a check that rootCategory >= recursiveCategory >= terminalCategory, and no levels are skipped between them.

candidategenerator.js
```diff
@@ -61,23 +129,29 @@ window.GEN = function(sTree, words, options){
...
 			recursiveOptions[k] = options[k];
 	}
 
-	var rootlessCand = addPhiWrapped(gen(leaves, recursiveOptions), options);
+	/* if rootCategory and recursiveCategory are the same, we don't want to call
+	 * addRecCatWrapped becasue half of the candidates will have a root node with
+	 * only one child, which will be of the same category, ie. {i {i (...) (...)}}
+	 */
+	var rootlessCand = gen(leaves, recursiveOptions)
```
jbellik: good catch

candidategenerator.js
maxTarlov: So this functionality is actually the issue I was running into with the exhaustivity option. In light of this, the code will actually just warn the user and won't mess with pCat

candidategenerator.js
maxTarlov: As it stands, the defaults for rootCategory and terminalCategory are based on the value of recursiveCategory. Is this the way we want it or do we want to do it some other way, eg. the defaults for recursiveCategory and terminalCategory could be based on rootCategory.

maxTarlov: Also: the default for recursiveCategory is "phi", do we want to make reference to the pCat instead? I feel like either way someone might get confused.

candidategenerator.js
maxTarlov: You are right, thanks for catching that

candidategenerator.js
maxTarlov: I've changed it to pRoot

candidategenerator.js
maxTarlov: GEN no longer manipulates pCat, as per your request, but this warning does show up in an alert box and in an error

candidategenerator.js
maxTarlov: As mentioned above, GEN can no longer modify pCat, a warning pops up and an error is thrown instead. These changes will be pushed later today

candidategenerator.js
maxTarlov: recCat replaced with pCat.nextlower(pRoot.cat)

candidategenerator.js
maxTarlov: updated in forthcoming commit

candidategenerator.js
jbellik: These are good questions. I think we should stick with hard-coded 'phi' as the default, just in case pCat changes somehow. Also, defining the default root and terminal categories based on the recursive category seems good to me.

candidategenerator.js
jbellik: Sounds good.

candidategenerator.js
```diff
@@ -41,6 +41,13 @@ var terminalNum = 0;
 window.GEN = function(sTree, words, options){
 	options = options || {}; // if options is undefined, set it to an empty object (so you can query its properties without crashing things)
 
+	/* First, warn the user if they have specified terminalCategory and/or
+	 * rootCategory without specifying recursiveCategory
+	 */
+	 if(!options.recursive && (options.rootCategory || options.terminalCategory)){
```
jbellik: This should be options.recursiveCategory, I believe

jbellik:
This will break existing analysis files that rely on requirePhiStem as a GEN option.

PR #131, opened 2019-10-07
adapt tableauMaker to accept options on constraints

jbellik:
Format: constraint-category-options
options needs to be a JSON string; use JSON.stringify({myoption: myvalue}) to make this. tableauMaker uses JSON.parse to turn it back into an object.


jbellik:
Yes, I omitted the options from the tableau header deliberately because
they made the rows ridiculously wide and it was kind of ugly. But it's an
open question whether we might actually need to show them. In fact, it
might be important to show them, in the interest of being able to replicate
the analysis based on the tableau it generates.

A possible compromise would be showing some kind of beautified version of
the options object: matchSP-xp-requireLexical, matchSP-xp-requireOvertHead.
Still gets really ugly if you set both to true:
matchSP-xp-requireLexical-requireOvertHead...

jbellik:
I changed things so the options do show up in the header.

PR #138, opened 2019-10-15
Interface Info Button and obeysHeadedness fix

maxTarlov:
Added info button for tone annotations.
fixed obeysHeadedness option so that it checks the whole tree, not just the root node
fixed built in analysis code to work with new tone options in interface
fixed interface tone options which had a small issue with the cache and reloading


built-in_analyses.js
```diff
@@ -15,14 +15,18 @@ function my_built_in_analysis(){
...
   for(var x = 0; x < toneButtons.length; x++){
+    toneButtons[x].parentNode.setAttribute("style", "display: table-cell");
     if(toneButtons[x].value="addIrishTones_Elfner"){
-      toneButtons[x].setAttribute("checked", true);
+      toneButtons[x].setAttribute("checked", "checked");
     }
     else{
       //we don't want multiple radio buttons to be checked, it gets confusing
-      toneButtons[x].setAttribute("checked", false);
+      //this isn't doing what it is supposed to, I don't know why -Max 10/10/19
```
jbellik: What is it supposed to be doing that it's not doing? It looks like this is intended to ensure that no other tone option can be selected, but with radio buttons, that's automatically the case.

jbellik: Ok, upon further review, I think I see the issue, which is that even though the html shows only one button selected, the objects still both have "checked":"checked". But I think the reason this isn't working is because the value of the Japanese radio button is getting reset

jbellik: Fixed it. See comment above on the conditional.

built-in_analyses.js
```diff
@@ -15,14 +15,18 @@ function my_built_in_analysis(){
   /* Step 4: (optional) If you want to annotate your tableaux with tones,
    * uncomment this block: */
   /*
-  var toneButtons = document.getElementsByName("toneOptions");
+  var toneCheckbox = document.getElementById("annotatedWithTones");
+  toneCheckbox.setAttribute("checked", "checked");
+  var toneButtons = toneCheckbox.parentNode.parentNode.getElementsByTagName("input");
   for(var x = 0; x < toneButtons.length; x++){
+    toneButtons[x].parentNode.setAttribute("style", "display: table-cell");
     if(toneButtons[x].value="addIrishTones_Elfner"){
```
jbellik: This needs to be a "===". You were resetting the value of all the other buttons to "addIrishTones_Elfner"! This is what was causing problems below.

built-in_analyses.js
```diff
@@ -37,17 +41,27 @@ function built_in_Irish(){
...
   //some stuff for tones
-  var toneButtons = document.getElementsByName("toneOptions");
+  var toneCheckbox = document.getElementById("annotatedWithTones");
+  toneCheckbox.setAttribute("checked", "checked");
+  var toneButtons = toneCheckbox.parentNode.parentNode.getElementsByTagName("input");
   for(var x = 0; x < toneButtons.length; x++){
+    toneButtons[x].parentNode.setAttribute("style", "display: table-cell");
     if(toneButtons[x].value="addIrishTones_Elfner"){
-      toneButtons[x].setAttribute("checked", true);
+      toneButtons[x].setAttribute("checked", "checked");
```
jbellik: I think true should've been an okay value as well, but "checked" is fine too.

built-in_analyses.js
```diff
@@ -37,17 +41,27 @@ function built_in_Irish(){
...
+
+  //document.getElementById("exhaustivityDetailRow").style.display = "block";
   //some stuff for tones
-  var toneButtons = document.getElementsByName("toneOptions");
+  var toneCheckbox = document.getElementById("annotatedWithTones");
+  toneCheckbox.setAttribute("checked", "checked");
+  var toneButtons = toneCheckbox.parentNode.parentNode.getElementsByTagName("input");
   for(var x = 0; x < toneButtons.length; x++){
+    toneButtons[x].parentNode.setAttribute("style", "display: table-cell");
     if(toneButtons[x].value="addIrishTones_Elfner"){
```
jbellik: Again, this needs to be "===" to check for strict equality rather than "=" which is doing assignment

candidategenerator.js
```diff
@@ -167,12 +167,42 @@ window.GEN = function(sTree, words, options){
...
+	}
+
+	//outer function
+	//first, check the parent node
+	if (!nodeIsHeaded(tree))
+		return false;
+	//return false if one of the children does not obey headedness
+	if (tree.children){
+		for (var x = 0; x<tree.children.length; x++){
+			if (!obeysHeadedness(tree.children[x])) //recursivce function call
```
jbellik: recursive

candidategenerator.js
```diff
@@ -167,12 +167,42 @@ window.GEN = function(sTree, words, options){
...
+		 */
+		var children = node.children;
+		//vacuously true if node is terminal
+		if (!children)
 			return true;
-	return false;
+		for (var i = 0; i < children.length; i++)
+			if (children[i].cat === pCat.nextLower(node.cat))
+				return true;
+			return false;
```
jbellik: Please add the curly braces for the for loop and if statement; right now the indentation  makes it look like "return false" is inside the loop, which it (correctly) isn't really.

candidategenerator.js
```diff
@@ -167,12 +167,42 @@ window.GEN = function(sTree, words, options){
...
 }
 
-function rootIsHeaded(pRoot, recCat) {
-	var children = pRoot.children || [];
-	for (var i = 0; i < children.length; i++)
-		if (children[i].cat === pCat.nextLower(pRoot.cat))
+/* Function to check if a tree obeys headedness. Each node must either be be
+ * terminal or have at least one child of the category immidately below its own
+ * on the prosodic hierarch. Otherwise, return false. Written as a recursive
+ * function, basically a constraint.
```
jbellik: Okay, this is nice implementation of a check for headedness at all levels. That's not quite what I had in mind (see issue description), but after discussing the expected use cases with Richard, I've concluded that this fix is fine and we don't need to worry about people wanting to enforce headedness at only some levels.

interface1.html
```diff
@@ -500,31 +500,26 @@ <h2>Options for prosodic tree generation (GEN function)</h2>
...
-				<i>Annotate trees with tones for...</i><br/>
-				<table class="spaceytable"><tbody><tr>
-					<td><input type="radio" name="toneOptions" value="noTones" checked="checked">None</td>
-					<td><input type="radio" name="toneOptions" value="addJapaneseTones">Japanese</td>
-					<td><input type="radio" name="toneOptions" value="addIrishTones_Elfner">Irish</td>
-					
-				</tr></tbody></table>
-				<hr/>
-				-->
+				<span class="info"><span class="content" style="width: 500px">
```
jbellik: Upon further reflection, I think it would be better if this info was split into 2 info sections, one info button that goes with Japanese and another that goes with Irish. They should appear and disappear when the "annotate trees with tones" option is checked / unchecked.

built-in_analyses.js
maxTarlov: I found that setAttribute() wasn't actually doing anything unless the second argument was a string. I could have used the string "true" instead of the value true, but the rest of the interface uses "checked" so I decided it would be good to keep that consistent.

interface1.html
maxTarlov: I can do that, but it really messes with the formatting once one of the info buttons is clicked. I'll send a commit with two separate info buttons so you can see what I mean.

built-in_analyses.js
maxTarlov: well there was a whole thing with setAttribute() so i've reverted back to toneButtons[x].checked = "checked";

interface1.html
maxTarlov: I've written a JS solution to the formatting issue trying to keep everything looking as similar as possible to the standard info button formatting and mechanics.

PR #144, opened 2019-10-23
Equal sisters bug fix, matchMin and constraint definitions

maxTarlov:
I though this was in another branch that already got merged, but it was in this branch, which had not been merged. It just removes the category options in the interface. Since equalSisters doesn't make any reference to the category, it doesn't matter if it is there or not for violations, but if it is there it will confusingly be in the tableaux headers.

I also added matchMinSP and matchMinPS since they weren't there before.

Finally, I moved all of the constraint definitions that existed as mouse-over boxes into info buttons. This is part of a larger project to precisely define all of the constraints in pros and cite sources better.


jbellik:
Somewhere in the merge process (probably bc of an update from master?), the info button update got lost.

PR #150, opened 2019-10-28
Match refactor

maxTarlov:
matchSP now takes the options maxSyntax, nonMaxSyntax, minSyntax, nonMinSyntax, maxProsody, nonMaxProsody, minProsody, nonMinProsody. The test file take a while to load, I tested every combination of these parameters (81 total match constraints)


test/matchOptions.html
```diff
@@ -0,0 +1,84 @@
...
+				return matchSP(stree, ptree, cat, {minSyntax: true, minProsody: true});
+			}
+			function minMax(stree, ptree, cat){
+				return matchSP(stree, ptree, cat, {minSyntax: true, maxProsody: true});
+			}
+			*/
+
+
+			//'matchPS-phi-{"maxProsody":true}'
+			var con = ['matchSP-xp', 'matchSP-xp-{"maxSyntax":true}', 'matchSP-xp-{"maxProsody":true}', 'matchSP-xp-{"requireOvertHead":true}', 'matchPS-phi-{"requireLexical":true}'];
```
jbellik: matchSP-xp-{"maxProsody":true} and matchPS-phi-{"requireLexical":true} are a bit strange as constraints go, so I'm having trouble deciding if they are "working". Please sanity-check my interpretations:
* matchSP-xp-{"maxProsody":true} = assign a violation for every xp that doesn't have a corresponding maximal phi. --> That does seem to be happening.
* matchPS-phi-{"requireLexical":true} = assign a violation for every phi that doesn't have a corresponding lexical XP. --> that also seems to be happening.

maxTarlov: these are the definitions I was working with

PR #152, opened 2019-10-29
inital reorganize. Gen options moved to top. Combined string input se…

ColinAChen:
…ctions. Removed tree builder section from bottom. Added divider lines with extra padding in between new sections. First input area should correctly be used for gen options.

### DIFF
--- a/interface1.html
+++ b/interface1.html
@@ -300,9 +300,10 @@
 					</span>
 				</span>
 			</h2>
+			<p>String of terminals: <input type="text" name="inputToGen"> <button id="goButton" type="button">Generate tree</button></p>
 			<button id="startTreeUIButton" class="orange-button" type="button">Open tree-builder</button>
 	            <div id="treeUI" style="display: none">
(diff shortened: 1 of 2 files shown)


ColinAChen:
The most recent commit should restore functionality to the Build Tree button

jbellik:
Much better. But it's still somewhat confusing in that clicking "Generate tree" doesn't automatically open the tree-builder. I'd actually like to eliminate the "Open tree builder" button altogether, and have the tree builder open when "Generate tree" is clicked.

ColinAChen:
The "open tree builder" button should be removed and the tree builder menu should open when "generate tree" is clicked.

jbellik:
Looking good. Can I ask for two more changes?
1. Rename "Generate tree" to "Build syntax"
2. Create another button next to it called "Run GEN" that does the same thing as the "Get results" button.
Thanks!

ColinAChen:
Both of the changes should be implemented!

PR #170, opened 2019-11-07
Drop down menu

ivpetkov:
Changed display of built in analysis to drop down menu. Currently the built_in_Irish() function is called when the selection in the drop down is changed. As more analyses are added, the onchange function will have to call a function that calls different function based on the value of the option in the dropdown.


ivpetkov:
I added the new built-in system to the menu.

jbellik:
This seems to basically be working, although it doesn't automatically clear the previous selections when a new one is made. But I think Max was going to write that into the functions that load the built-in systems, so that's fine.

jbellik:
Am I right in thinking that this menu assumes that the function for every built-in system will be named built_in_[system name]()? That seems fine but we should document it in built_in_analysis.js.

PR #174, opened 2019-11-07
update interface-reorg from master

jbellik:
Most of these changes have nothing to do with the interface, but a couple touch the text and one adjusts the css.


jbellik:
Actually nevermind, this will probably overwrite existing changes to the interface...

PR #176, opened 2019-11-08
Scrollable tree builder window and done button message

ivpetkov:
Tree builder is scrollable, resizable, displays one tree per row, and snaps to bottom of window when a new tree is created.


ivpetkov:
I also added the text to show a visual signal that the "Done!" button worked to this branch. When the user clicks done, the message appears and when the user clicks "Build syntax" the message disappears.

ivpetkov:
I updated the done message so that is also disappears when the user edits a tree by clicking "Add mother" or "Delete".

PR #185, opened 2019-11-18
Interface reorg

ColinAChen:
sub-constraints show/hide on line below constraint


jbellik:
We don't want to merge into master until the entire interface has been redone.

PR #188, opened 2019-11-20
Save load analysis

maxTarlov:
Save/load functionality


built-in_analyses.js
```diff
@@ -28,48 +28,40 @@ function clearAnalysis(){
...
+          }
           //open the fieldset
-          conFields[x].setAttribute("class", "open")
-          //reveal the categories
-          con_trs[y].setAttribute("class", "constraint-checked")
-          //iterate over the check boxes for cateogories
-          //assumes that the constraint is the first check box
-          for(var z = 1; z < con_boxes.length; z++){
+          conFields[x].setAttribute("class", "open");
+          console.log(x);
```
jbellik: should probably comment this out

maxTarlov:
Changes integrated:
* added line break in "save" message
* save message changes when things in the analysis change
* file name is not entered by user in a pop-up window
* "upload an analysis" disappears when an analysis is loaded and when "clear all" is pressed

PR #191, opened 2019-11-21
Gen warnings

ColinAChen:
Gen warnings for long strings and clitic movement


ColinAChen:
check pString split on space

Issue #98, opened 2019-09-05
Match options: add to interface


jbellik:
Recently, options were added to the match function to enable Match to ignore nodes marked as "silentHead: true" or "func: true" (i.e., only look at nodes with overt heads or lexical heads). The interface doesn't have any functionality to make these accessible, however.

The interface should also make it possible to select subcategories for the syntactic/prosodic categories that are selected -- i.e., Match(XP-max, phi-max), Match(XP-max, phi), Match(XP-nonmin, phi-nonmin), etc. These options haven't yet been built into the match function proper, but are coming soon.

jbellik:
This is going to be a large-ish addition to the interface, but I think you are equipped to tackle it because we already hashed out the design aspect of it. 
See: 
* constraints/match.js for the comments on matchSP that explain what all the options are
*  test/constraint_options_test.html for how to format the options for passing them into make_tableau()

You'll need to change the spotForm.onsubmit function in interface1.js to deal with options for the constraints. There isn't yet support in the interface for any kind of constraint options.

Let me and/or Max know if you have trouble figuring out what interface settings are supposed to go with which matchSP() options. I think the names are fairly transparent though.

jbellik:
Match(Syntax):
- all syntactic nodes -> only need options for requireLexical and requireOvertHead
- all maximal syntactic nodes -> {maxSyntax: true}
- all non-minimal syntactic nodes -> {nonMinSyntax: true}
- custom: this is the only place where you might need to deal with the other options (maxProsody, minProsody, minSyntax, etc.)

jbellik:
For all the versions of MatchSP: 
if the user selects "lexical" --> {requireLexical: true}
if they select "overtly headed" --> {requireOvertHead: true}
if they select both --> {requireLexical: true, requireOvertHead: true}

If you don't specify the value of one of the options, it will evaluate to false, so there's no need to explicitly set any of them to false. But it also doesn't do any harm to set options to false if the checkbox wasn't selected.

PR #198, opened 2019-11-24
Adding match options to interface

ivpetkov:
Completes issue #98. I think all the match options are working correctly. I changed the text for custom match from the wireframe to say "Enforce Match only for syntactic categories that are..." instead of "Enforce Match only for XPs that are..." because I couldn't figure out how to update the text based on which categories are checked.


maxTarlov:
I changed the wording in the interface just a little bit to try to improve clarity, but mostly I just added info buttons for all of the options so it is easy to find what we mean by, e.g. "only lexical nodes." I had some trouble getting the info buttons for the options under the custom match constraint to indent, @ivpetkov can you make that happen? (I really don't know any CSS).

ivpetkov:
I'll fix the indentation.

Issue #195, opened 2019-11-21
NoShift error


jbellik:
Convert noShift error in the console to a warning pop-up. Catch the error instead of causing the function to just crash.

And/or add a warning whenever no syntactic tree is entered?

jbellik:
Iva, can you get this one? Should be a quick fix in constraints/noShift.js

ivpetkov:
Yep, I'll start working on it!

PR #213, opened 2019-12-05
noShift error alert

ivpetkov:
Added alert message with noShift error message that is triggered when error is thrown. Should the error be somehow handled when it is caught, besides displaying the alert?


ivpetkov:
Issue #195

jbellik:
It's kind of unfortunate that this now results in the user having to click "Ok" to dismiss the pop-up once for every candidate that's being generated. Maybe console.warn() would be better than the pop-up I originally proposed. If the user forgot to put in a syntactic tree, there's not much to be done about that.

Issue #230, opened 2019-12-06
Add AlignPS to interface


jbellik:
They should go under the existing Align constraints, like so:
AlignLeft(syntax -> prosody)
AlignRight(syntax->prosody)
AlignLeft(prosody-> syntax)
AlignRight(prosody-> syntax)

Functions are defined in constraints/align.js

PR #236, opened 2019-12-08
added alignPS to interface

ivpetkov:
Issue #230


ivpetkov:
What should I put for the constraint definitions? They aren't in the google doc.

jbellik:
Good question. I wrote some and added them.

PR #243, opened 2020-01-09
it was a scope issue

maxTarlov:
"options" got initialized twice in makeTableau somehow. I have no clue why this only affected the built-in systems, but the tones are all working now for built-in systems and for non-built-in systems.


maxTarlov:
Update on cause: it was never an issue with the built-in analyses, I think it was an issue with the constraints that the built-in analyses used. The second "options" variable didn't get initialized unless at least one of the constraints had an options object. All of the match constraints have options objects even if they are empty (matchSP-xp-{}, e.g.), so it was match constraints, not the built-in systems that used them, which caused this problem.

Issue #239, opened 2019-12-09
Tableaux accumulation


jbellik:
Add a button to clear previous tableaux on interface?

PR #244, opened 2020-01-14
added button to clear previous tableaux

ivpetkov:
Issue #239 
Is button placement and styling ok?


jbellik:
Styling and functionality look good. Placement seems a bit odd. Maybe place it next to "Get results" instead, and call it "Clear results" for symmetry.

Issue #245, opened 2020-01-14
"String of terminals" overriding tree terminal strings?


nkalivoda:
If you have something in "String of terminals" when you hit "Done!" under the tree-builder, it uses that to generate the outputs for all of the inputs, which is a problem. Maybe we could have "String of terminals" automatically emptied when you hit "Done!".

PR #251, opened 2020-01-21
'string of terminals' is cleared when you click 'done'

ivpetkov:
I cleared of the "String of terminals" when the user click the "Done" button. Does this solve issue #245? 

I don't completely understand the bug, so I couldn't reproduce the problem, I just implemented Nick's solution of having "String of terminals" automatically emptied when you hit "Done!"


jbellik:
As you pointed out during the meeting last week, this doesn't fix the problem. We need to be checking whether there are any trees in the tree-builder before using the string of terminals at all.

Issue #137, opened 2019-10-15
Make Match conflate sCats with identical terminals


nkalivoda:
Elfner's definition of Match treats [[...xyz...]] as a single sCat, since the two dominate the same terminals. That is, failure to match [[...xyz...]] results in 1 violation, not 2.

We shouldn't expect users to know that they need to pre-trim their syntactic trees. They might put in [[...xyz...]] expecting Match to already work like Elfner says. So can we change the definition of Match to do this?

jbellik:
I wonder if a better option would be to create a function that trims the strees, rather than bundling this into Match. Is relevant for Align as well as Match, for instance? Would we ever want to give [[...xyz...]] two violations?

nkalivoda:
It will generally be relevant for Align too, but only because alignment analyses assume the Lexical Category Condition, which ensures that [[...xyz...]] will never arise. The bigger sCat here would be invisible by the LCC––not because it shares the terminal string of the smaller sCat, but because it has a silent head (presumably a trace or a functional projection; the LCC doesn't mention silent in situ lexical heads, since the issue never comes up, but this might be an exception).

I like the tree-trimming idea. We should do that instead of modifying Match. However, I think that:

(1) Inputs should still display with the full original bracketing from the user's sTree, even when nodes have been trimmed for Eval.
(2) It might be nice to have an option to turn this off (perhaps not on the interface, but possible for those in the know), just in case someone ever wanted to experiment with a different definition that actually counts both sCats in [[...xyz...]]. But this is really less important, and not interface-button-worthy.

nkalivoda:
Also, this only applies to sCats with the same category. So two phis with the same terminal string, but not an iota and a phi with the same terminal string.

jbellik:
Write 2 functions:
1. Take a syntactic tree and trim out all heads marked silent:true
2. Take a (trimmed) syntactic tree and trim out XP-XP configurations where both XPs have all and only the same terminals

jbellik:
This function has been implemented and merged into master. However, we need to add it to the interface as an option.

jbellik:
@nkalivoda , I am thinking we could add  this in either of 2 places on the interface:
1. under inputs -- since it has to do with how inputs are interpreted
2. under mapping constraints -- it's not a constraint but it's an option that is only relevant to the mapping constraints.

Which do you think is better? Please weigh in so Max can implement.

nkalivoda:
I'd say under the inputs. The more prominent the better.

PR #266, opened 2020-01-27
S cat conflation

maxTarlov:
Re: Issue #137 
This is the back-end for the tree trimming functionality. Adds 4 global functions:
* copyTree(): makes a copy of a tree's values in a separate reference so that the function other don't delete material from the input
* trimSilentTerminals(): removes silent terminals from the tree
* trimDeadEndNodes(): removes non-x0 terminal nodes that might have been created by the previous function if all of their children were silent
* trimRedundantNodes(): removes needless recursion such as [[arbitrary terminals]]. This might be in the input, but it might also have been made by the previous two functions.

I also implemented the tree trimming functionality into matchSP as an option that defaults to true. Should I do this for align as well?


constraints/match.js
```diff
@@ -112,10 +112,15 @@ function matchPS(sTree, pParent, pCat, options)
 *	minProsody: If true, the prosodic match needs to be minimal. Passed to hasMatch.
 *	nonMaxProsody: If true, the prosodic match must be non-maximal. Passed to hasMatch.
 *	nonMinProsody: If true, the prosodic match must be non-minimal. Passed to hasMatch.
+* trimTrees: Default true, trims trees of silent nodes, non-x0 terminals and redundant nodes, e.g. [[some terminals]]
```
jbellik: I think this should default to false, actually. Also we should make sure that the tableau displays whether trimming has taken place or not.

jbellik:
My other concern is what this function is doing to ptrees if you run MatchPS...

constraints/match.js
maxTarlov: ok, updated in commit 6ddcac2

maxTarlov:
new changes:
* trimTrees defaults to false
* matchPS deals with trimTrees so matchSP doesn't trim pTrees
* added the parenthesized trimmed tree to the tableaux when constraints use the trimTrees option

maxTarlov:
trimming of trees is now only an option in makeTableau

test/treeTrimming.html
```diff
@@ -0,0 +1,160 @@
+<!--Max Tarlov testing tree trimming 10/23/19-->
```
jbellik: This has obviously been updated since October.

jbellik:
@nkalivoda I'm going to merge this into master unless you have any objections.

PR #267, opened 2020-01-27
Overriding terminals

ivpetkov:
I kept the "done" button clearing "string of terminals", but now if trees have been added to the analysis, then gen does not take whatever is in "string of terminals" as input

### DIFF
--- a/interface1.js
+++ b/interface1.js
@@ -394,6 +394,13 @@ window.addEventListener('load', function(){
 		//Get input to GEN.
 		var pString = spotForm.inputToGen.value;
 
+		// Get the code that is in the stree textarea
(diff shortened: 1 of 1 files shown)


ivpetkov:
@jbellik this is the pull request where I thought i fixed issue #245

jbellik:
I figured it out: this is another change that got overwritten by a feature branch @ColinAChen  was working on. I think it's the same one that we were struggling to figure out last week.
Here's the commit that undid this fix: https://github.com/syntax-prosody-ot/main/commit/b9dc18d6b94c71e7c6552faa29724a31f830fcb3#diff-94a489a20c3cf3d811ed7d8313e25e88

Issue #252, opened 2020-01-22
hyphens (or underscores or commas?) rather than spaces between morpheme names in AlignLeftMorpheme


rbibbs:
Currently when you specify more than one morpheme name for the AlignLeftMorpheme constraint, spaces will be put between them in the constraint name in the output csv file. However, to correctly run a property analysis with OTW, none of your constraint names can have spaces. Therefore, we should probably try and put some other character than spaces to separate the morpheme names.

jbellik:
@rbibbs , please try putting ''' ''' (or whatever the Excel string delimitor is) around the constraint name and see if it still breaks the PA in OTWorkplace. Also try ; and ,.

Lexical item(s): a b --> this will break OTWorkplace
Lexical item(s): a-b --> I believe this will not break OTWorkplace

The user should still be able to put in space separated inputs on the interface, but SPOT should eliminate spaces from the tableau output.

rbibbs:
I'm unsure what the Excel string delimiter is, but semicolons (;) work great without it.

Another possibility I've though of: what if we just make the file type .tsv rather than .csv? Wouldn't we then be able to use commas? Not sure if this is a bigger ask than just changing how it's written.

PR #268, opened 2020-01-28
replaced space with underscore for alignLeftMorpheme category

ivpetkov:
issue #252


ivpetkov:
I'll change it from underscore to semicolon

PR #270, opened 2020-01-29
instructions on how to mark nodes as functional and have silent heads…

ColinAChen:
… in their respective custom match constraint options. Please let me know if this information should appear somewhere else.


jbellik:
Please also add this information to the tree-builder info section.
Also, will it be a problem if they put ", func" or ", silentHead" (with a space)? I think we should make this capable of handling spaces.

ColinAChen:
I added the info to the tree builder info section. It should already be able to handle any number of spaces so long as they are separated by commas.

PR #271, opened 2020-01-29
interface.js to add func and silentHead attributes to nodes. treeForm…

ColinAChen:
…atting parenthesize trees returns toneTree with func/silentHead/other attributes that are not id/cat/children. I am not sure if this is the correct list.


ColinAChen:
The changes to interface.js are identical to the ones in functional-silent-nodes branch so if these changes are correct, then you just need to merge this branch

jbellik:
Please comment here when you've made these changes so I know to look at the code again. If at all possible, please finish this by noon Tuesday so that I can review the code before the meeting on Wednesday. Thanks!

ColinAChen:
image
Is this closer to what is it is supposed to be doing?

jbellik:
Yes, much better! Looks good!

treeFormatting.js
```diff
@@ -33,14 +33,95 @@ function parenthesizeTree(tree, options){
...
+		// 			else{
+		// 				console.log("none");
+		// 				parTree.push(parens[node.cat][0]);//pushes the right parens
+		// 			}
+		// 		}
 		if (nonTerminal) {
 			if (visible) {
-				parTree.push(parens[node.cat][0]);//pushes the right parens
+				if (node["func"] && node["silentHead"]){
+					console.log("both");
```
jbellik: Please take out these "console.log"s and probably also the large commented sections above before the final merge. Thanks.

jbellik:
Thanks. One more thing: please mark terminal nodes with .f and .sh as necessary as well.
For the non-terminals, I don't think the . is necessary, although it's not really a problem. [f and [sh --> fine, [.f and [.sh --> also ok.

e.g. a tree [a [b]] where b = {cat: w, id: b, func: true, silentHead: true} should get linearized as [ a [ b.f.sh ]]

Does that clarify?

maxTarlov:
Hi everybody,

I was working on the head marking for the Japanese and the marking of heads in tree parenthesization and I decided that, since this branch had changes to treeformatting.js, I should merge those changes into my branch. Long story short I think I merged my japaneseHeadMarking branch into this branch in the process of resolving a merge conflict and now I am not totally sure what to do. I tried reverting the merge, but it looked like my headMarking code was still there.

I'm really sorry if this made things more complicated for you,

Max

jbellik:
Colin, I need you to comment here when you've made changes. And in general, I do need you to reply to comments and emails so that I know you've seen them. It can just be a quick "Will do" or "Can you clarify X?" or "I think Y will be a problem." Thanks.

ColinAChen:
I just pushed. Let me know if it is doing what you expect. Sorry for the lapses in communication, I'll be more on top of it in the future.

jbellik:
I can't merge this pull request because something's gone wrong in the version history. We need to relocate these changes to a new branch and make a new pull request.

jbellik:
Nevermind, @ozanblk showed me how to cherry pick the relevant commits from this branch. See new-parenthesize-tree. No need to redo anything.

PR #280, opened 2020-02-02
Built in system updates

maxTarlov:
* updated ICPP built-in analyses so GEN produces trees that are rooted in phi; apparently I had already added that functionality to the built-in analyses but had forgotten about it.
* updated built-in systems to load options with a drop-down selector
* updated clearAll() to clear options with a drop-down slector
* updated save analysis to save options with a drop-down selector

### DIFF
--- a/built-in_analyses.js
+++ b/built-in_analyses.js
@@ -101,11 +101,26 @@ function built_in_con(input){
       if(optionBoxes.length){
         //iterate over option checkboxes corresponding to this input
         for(var x in optionBoxes){
-          if(input[i].options[optionBoxes[x].value]){
-            optionBoxes[x].checked = true;
+          //dealing with checkboxes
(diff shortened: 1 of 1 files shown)


built-in_analyses.js
```diff
@@ -256,7 +287,7 @@ function built_in_Japanese_IM2017(){
 
 //cf. analysis_html_files/abstractMatchAnalysis.html. Japanese rebracketing project, Kalivoda 2019.
 function built_in_Japanese_rebracketing(n){
-  var gen = {obeysExhaustivity: true, requireRecWrapper: true};
+  var gen = {obeysExhaustivity: true, requireRecWrapper: true, rootCategory: "phi"};
```
jbellik: Ah, I see what happened. requireRecWrapper doesn't work on the interface, only on the backend.

PR #282, opened 2020-02-04
Input gen

jbellik:
Creates a function to make input syntactic trees using the existing GEN function. 

To include only binary-branching trees, this function *filters* the output of GEN, rather than not making the ternary+ branching trees in the first place, but it's fine for short strings.


jbellik:
Based on our conversations today, I am merging this branch even though there aren't any reviews on Github.

PR #289, opened 2020-02-08
Bug Fix

maxTarlov:
The tableau maker was leaving out the first candidate from GEN every time.


maxTarlov:
You can see the bug in the current version of Master by going to the interface, running GEN on a string of any length and then, in the console, parenthesizing the first parse from GEN for the same output and searching the page for that tree.

jbellik:
Ooops, thanks for catching that, Max!

PR #297, opened 2020-02-12
Japanese head marking

maxTarlov:
Gives nodes the property "isHead" when they are heads as defined for the Japanese system. In parenthesize tree, gives heads an astrisk marking.

* I made it so that only-child terminals don't get an asterisk, although they still evaluate as being a head. I did this because they are clearly heads, though I can easily remove this feature.

* Note that, because GEN re-uses elements, the tableau maker will need to run markHeadsJapanese for each pTree _right_ before parenthesizing it. After the sCat conflation confusion, I am unsure how we want to implement this in makeTableau

### DIFF
--- a/treeFormatting.js
+++ b/treeFormatting.js
@@ -54,11 +54,11 @@ function parenthesizeTree(tree, options){
 					parTree.push(parens[node.cat][0]);
 				}//pushes the right parens}
 
-				if(node["isHead"]){
+				if(node.isHead === true){
 					parTree.push('*'); //marks head with a *
 				}
 
-				if(node["isHead"] || node["func"] || node["silentHead"]){
+				if(node.isHead === true || node["func"] || node["silentHead"]){
 					parTree.push(' '); //push a space
 				}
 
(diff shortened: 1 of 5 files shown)


jbellik:
* binMaxHead should call this directly (@ivpetkov )
* marking heads should be an options on makeTableau()
* go ahead and mark only child heads with * as well

maxTarlov:
* marking only-child heads again
* added an option in makeTableau (and in parenthesize tree) for showHeads.

treeFormatting.js
```diff
@@ -33,30 +34,35 @@ function parenthesizeTree(tree, options){
...
+					parTree.push(parens[node.cat][0] + ".f");
+				}
+				else if (node["silentHead"]){
+					parTree.push(parens[node.cat][0] + ".sh");
+				}
+				else{
+					parTree.push(parens[node.cat][0]);
+				}//pushes the right parens}
+
+				if(node.head && options.showHeads){
```
jbellik: This is pushing a * right after the opening parenthesis. We actually want it after the closing parenthesis.
Ex:
{(* a b c*)} should be {(a b c*)*}
(* (a b*) (* c*)) should be ( (a b*) (c*)* )

PR #300, opened 2020-02-13
starCat constraint

nkalivoda:
Creates starCat constraint and corresponding test file. Needed for built-in systems from Truckenbrodt, which I plan to make. Should go somewhere on the interface, but I'm not sure where.


constraints/starCat.js
```diff
@@ -0,0 +1,15 @@
+//Assign a violation for every node of category c in p.
+//Truckebrodt (1995, 1999): *phi
+function starCat(s, p, c){
+	var occurances = 0;
+	if (p.cat === c){
+		occurances ++;
+	};
+	if (p.children){
```
jbellik: change to if (p.children && p.children.length) to be sure to handle the case where p.children has been declared but not initialized.

PR #311, opened 2020-02-21
S cat conflation interface

### LINKED ISSUE #301: Terminals with unrecognized categories disappear in stringification
In the parenthesizeTree function, ids are only pushed to the stringified tree if they have a corresponding entry in options.parens. So for example if you put in a tree where one of the terminals has category "y", it will completely disappear from the parenthesization of the s-tree, because the variable "visible" will get set to "false" for it.

This is confusing and should be changed.

### LINKED ISSUE #310: Tree trimming bug URGENT
Using the tree trimming with clitics results in an uxpected relabeling (or pruning problem) involved with the clitics.

Example:  

[[a [b]]   **[x.clitic]**] 

trimmed: [[a [b]] **dp.xp**]

We want the XP phrase boundaries around the clitic to be maintained through the trimming.
Might be caused by the "clitic" label on the XP in the input syntax, not really sure...

maxTarlov:
__Summary__:
* fixed #310 (clitics were being trimmed because the function trimmed non-x0 terminals)
* fixed #301 (changed showNewCats option default to true, since trees can be trimmed now)
* added a checkbox and info button for tree trimming (#137)

__Questions__:
* Do we want to only trim terminals of the categories "cp" and "xp" instead of all categories other than "x0" and "clitic"?
* Do we want unrecognized categories to show up in the trimmed tree stringification? I could also remove them from the tree object itself, but the mapping constraints ignore them either way.

### DIFF
--- a/interface1.html
+++ b/interface1.html
@@ -543,7 +543,12 @@
 			<p>String of terminals: <input type="text" name="inputToGen"> <button id="goButton" type="button">Build syntax</button><button class="orange-button" type="submit" style='margin-left: 5px'>Run GEN</button></p>
 
             <div id="treeUI" style="display: none">
-								<div class="constraint-row"><input type="checkbox" name="genOptions" id="trimTrees">Trim Syntactic tree</div>
+								<div class="constraint-row">
+									<span><input type="checkbox" name="genOptions" id="trimTrees">Trim Syntactic tree</span>
+									<div class="info">
+										<span class="content">When checked, evaluates mapping constraints against a version of the input tree where silent nodes and terminals of a category other than "x0" or "clitic" are removed. Also removes vaccuous recursion, eg. [[a b]]</span>
+									</div>
+								</div>
 
               	<div id="treeUIinner" style="display: none">
               		<hr/>
--- a/interface1.html
+++ b/interface1.html
@@ -667,7 +667,7 @@
 							<div class="constraint-row">
 								<span><input type="checkbox" name="constraints" value="wrap">Wrap</span>
 								<div class="info">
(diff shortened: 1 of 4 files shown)


interface1.html
```diff
@@ -536,6 +543,12 @@ <h2>Inputs
 			<p>String of terminals: <input type="text" name="inputToGen"> <button id="goButton" type="button">Build syntax</button><button class="orange-button" type="submit" style='margin-left: 5px'>Run GEN</button></p>
 
             <div id="treeUI" style="display: none">
+								<div class="constraint-row">
+									<span><input type="checkbox" name="genOptions" id="trimTrees">Trim Syntactic tree</span>
+									<div class="info">
+										<span class="content">When checked, evaluates mapping constraints against a version of the input tree where silent nodes and terminals of a category other than "x0" or "clitic" are removed. Also removes vaccuous recursion, eg. [[a b]]</span>
```
jbellik: - Cite Elfner 2012. Something like: "in accordance with the prosodic visibility theory proposed in Elfner (2012)"
- vacuous, not vaccuous
- How do nodes get marked as silent?
- Specify that both trees will be shown in the tableau: "The tableau will display both the original and the pruned versions of the syntactic tree."

interface1.html
```diff
@@ -536,6 +543,12 @@ <h2>Inputs
 			<p>String of terminals: <input type="text" name="inputToGen"> <button id="goButton" type="button">Build syntax</button><button class="orange-button" type="submit" style='margin-left: 5px'>Run GEN</button></p>
 
             <div id="treeUI" style="display: none">
+								<div class="constraint-row">
+									<span><input type="checkbox" name="genOptions" id="trimTrees">Trim Syntactic tree</span>
```
jbellik: Let's change this text to say "Prune phonologically-contentless syntactic nodes"

PR #315, opened 2020-02-25
Align morpheme

ColinAChen:
Align Right Morpheme implemented. I couldn't figure out how to use the test.html but I tested it through the interface and it seems to work.


ColinAChen:
I reimplemented with a generalized alignMorpheme that alignRightMorpheme and alignLeftMorpheme call

interface1.js
```diff
@@ -345,6 +345,7 @@ window.addEventListener('load', function(){
 			var constraintBox = spotForm.constraints[i];
 			if(constraintBox.checked){
 				var constraint = constraintBox.value;
+				console.log(constraint);
```
jbellik: Do we still need this console.log?

PR #316, opened 2020-02-25
exhaustivityParent implemented and tested through interface

ColinAChen:
I implemented the constraint but have not added it to the interface.


constraints/exhaustivityParent.js
```diff
@@ -0,0 +1,37 @@
...
+	
+	//Recursive case: if parent is non_terminal, find out if there are any violations in each of the subtrees rooted in its children.
+
+	if(ptree.children && ptree.children.length){
+		var vcount = 0;
+		var child;
+		for (var i=0; i < ptree.children.length; i++){
+			child = ptree.children[i];
+			if (ptree.cat!==child.cat && pCat.nextLower(ptree.cat)!==child.cat){
+				vcount++;
```
jbellik: This is incrementing vcount once for every level-skipping child. It needs to only increment it once per parent. I.e., break once you increment vcount the first time. Or use a while loop.

ColinAChen:
I reimplemented exhaustParent, renamed exhaust1 to exhaustChild in the constraint file and interface, and added exhuastParent to the exhaustivity.js constraint file.

interface1.html
```diff
@@ -1074,9 +1074,9 @@ <h3>StrongStart</h3>
 
 					<div class="constraint-selection-table">
 						<div class="constraint-row">
-							<span><input type="checkbox" name="constraints" value="exhaust1">Exhaustivity</span>
+							<span><input type="checkbox" name="constraints" value="exhaustChild">Exhaustivity</span>
```
jbellik: We will actually want to have exhaustParent as the version of exhaustivity that's visible on the interface by default. 
exhaustChild should go in the "show more" section.

constraints/exhaustivity.js
```diff
@@ -23,8 +23,36 @@ function exhaust1(s, ptree){
...
+		//var vcount = 0;
+		var child;
+		for (var i=0; i < ptree.children.length; i++){
+			child = ptree.children[i];
+			if (ptree.cat!==child.cat && pCat.nextLower(ptree.cat)!==child.cat){
+				//vcount++;
+				return 1;
+			}
+			if (exhaustParent(s,child) > 0){
+				return 1;
```
jbellik: Here we still need to be accumulating violations. I think it should be vcount++

constraints/exhaustivity.js
```diff
@@ -23,8 +23,36 @@ function exhaust1(s, ptree){
...
+	//Recursive case: if parent is non_terminal, find out if there are any violations in each of the subtrees rooted in its children.
+
+	if(ptree.children && ptree.children.length){
+		//var vcount = 0;
+		var child;
+		for (var i=0; i < ptree.children.length; i++){
+			child = ptree.children[i];
+			if (ptree.cat!==child.cat && pCat.nextLower(ptree.cat)!==child.cat){
+				//vcount++;
+				return 1;
```
jbellik: By using return here, we'll never assess exhaustivity violations in the subtrees when there is a violation higher in the tree. That's a problem.

ColinAChen:
I added an array to keep track of categories that have already been violated at each level so they don't get counted again.

constraints/exhaustivity.js
```diff
@@ -23,8 +23,34 @@ function exhaust1(s, ptree){
...
+	//Recursive case: if parent is non_terminal, find out if there are any violations in each of the subtrees rooted in its children.
+
+	if(ptree.children && ptree.children.length){
+		var vcount = 0;
+		var child;
+		alreadyViolated = [];
+		for (var i=0; i < ptree.children.length; i++){
+			
+			child = ptree.children[i];
+			if (ptree.cat!==child.cat && pCat.nextLower(ptree.cat)!==child.cat && !alreadyViolated.includes(child.cat)){
```
jbellik: This isn't quite right either, because there could be, for example, multiple phis that contain clitics. You could change this to push the id of the parent instead of the category, but I think it would be more efficient to use a while loop with a boolean flag. Take a look at the implementation of eqSis and strongStart in equalSisters.js and strongStart.js respectively.

constraints/exhaustivity.js
```diff
@@ -44,13 +44,12 @@ function exhaustParent(s, ptree){
 	if(ptree.children && ptree.children.length){
 		var vcount = 0;
 		var child;
-		alreadyViolated = [];
```
jbellik: You deleted your variable declaration...

PR #317, opened 2020-02-25
custom align in interfrace with same option as custom match

ColinAChen:
I added a custom align constraint in the interface. I have not made any backend changes to support a custom align constraint.

Are these the right set of constraints? I copied them directly from the custom match constraint but the align constraints seem to take fewer arguments than the match constraints. 

If these are the right constraints, should I try implementing the custom align constraint?

### DIFF
--- a/interface1.html
+++ b/interface1.html
@@ -794,6 +794,87 @@
 								<div class="category-selection-div"><input type="checkbox" name="category-alignRight" value="x0">X<sup>0</sup></div>
 							</div>
 						</div>
+
+						<div class="constraint-selection-table">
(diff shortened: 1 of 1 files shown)


jbellik:
Hi Colin,
Yes, please implement the backend changes. I actually didn't mean to have you add this to the interface yet; I realize now I didn't take a close look at the issue when we assigned it. I will update it now to make it more clear & specific! Sorry about the unclarity. Let's actually reverse those changes to the interface because Iva is working on the constraint reorganization, and I would prefer to avoid a merge conflict.
Thanks.

PR #324, opened 2020-03-03
Focus marking

ColinAChen:
Mark nodes as focus nodes with ,foc. nodes are labeled as focus nodes with .foc in the tableau.


treeFormatting.js
```diff
@@ -90,18 +100,30 @@ function parenthesizeTree(tree, options){
...
+			if (node["silentHead"]){
+				tempLabel += ".sh";
 			}
-			else{
-				parTree.push(node.id);
+			if (node["foc"]){
+				tempLabel += ".foc";
 			}
+			parTree.push(tempLabel);
+			// if (node["func"] && node["silentHead"]){
```
jbellik: Can all these comments be deleted?

PR #325, opened 2020-03-03
Custom align

ColinAChen:
custom alignSP and PS. based on MatchPS and PS. It is added to the interface although I haven't tested it yet.


constraints/alignWrap.js
```diff
@@ -65,7 +108,10 @@ function alignLeftPS(sTree, pTree, cat){
 function alignRightPS(sTree, pTree, cat){
 	return alignPS(sTree, pTree, cat, 'right');
 }
-
+function alignCustom(sTree, pTree, cat, options){
+	options = options || {};
+	return alignSP()
```
maxTarlov, line 113: Shouldn't this return alignPS(sTree, pTree, cat, options)? This looks like it will always return 0 since no arguments are passed

jbellik:
Please add a test file on the model of matchOptions.html (@maxTarlov )

jbellik:
- It looks like the vcounts in the first 3 tableaux in the test file are fine.
- In the 4th tableau, the one that has input = {four [terminal]}, nonMaxSyntax does not seem to be evaluating correctly. It looks like [terminal] is being treated as nonMaximal when it shouldn't be.
- [six [terminal]] and [terminal seven] : [terminal...] is minimal in both so the 0 violations are correct
- {eight [terminal]}: looks like [terminal] is being treated as nonminimal?
- [nine [terminal]]: violations look correct
- [ten [terminal]] violations do not look right at all (all 0s)

jbellik:
I fixed most of the incorrect 0 results with my last commit (moved the closing bracket for walkTree() up 2 lines), but now there are some incorrect false positives happening -- things that should count as matches but are getting violations. See: https://docs.google.com/spreadsheets/d/1FDkG4QCKJyWcM0S7LZamyjrO23_rE9S2T8wrFnaFhOA/edit#gid=0

jbellik:
FURTHER UPDATE: I checked the call to makeTableau in the actual test file, and the remaining errors were due to the wrong trees being used in the candidate sets (i.e., prosodic candidates were being checked against different syntactic trees from the ones being shown in the lefthand corner of the tableau). All the tableau seem to be correct now.

jbellik:
@ColinAChen , please go through and double-check that all the tableaux now match the expected values, and then comment here and I will merge this pull request.

ColinAChen:
The customAlign test file looks good to me!

maxTarlov:
but now the short test file shows issues. It isn't assigning any violations for alignSP constraints

jbellik:
Shoot. Were the violation counts in the short test file working before?

jbellik:
Fixed it. Options hadn't been added to the alignLeft and alignRight functions!!!

jbellik:
(well, not quite fixed, but partially)

maxTarlov:
to answer your earlier question, violations were being assigned, but not correctly before your fixes

maxTarlov:
ok, now its working

jbellik:
Function works but interface needs work:
- Custom AlignLeftSP
- Custom AlignRightSP
- both of those should go in the "show more" section immediately after the AlignPS constraints
- set up alignCustom() function in the backend

jbellik:
Also need:
- custom AlignLeft / Right PS
- "match only" should "align only" in lexical option info

jbellik:
@ColinAChen , let's just delete the interface changes from this branch, as per the discussion on #370

PR #326, opened 2020-03-05
Multiple clitic movement

maxTarlov:
New function uses Heap's algorithm to generate permutations of the word list by swapping pairs of words. If clitics are specified, function will only swap pairs of words where at least one member is a clitic. If permuteAll is set to true, will generate all permutations.


jbellik:
Is there a test file for this?
I may not be able to review this before the meeting tomorrow -- sorry about the delay.

maxTarlov:
Moved relevant source code to new branch and opened fresh pull request #368 because of merge conflicts and lack of understanding of git

PR #328, opened 2020-03-09
Clitic problems

ivpetkov:
Issue #285 
I'm not sure if this is how you wanted the warning for using hyphens or underscores in category and id names to work, let me know if it should be different or what the message should be.


jbellik:
* allow underscores
* disallow hyphens 
* if a hyphen is found, then don't add trees to analysis or display the "done" message

jbellik:
* in the dialog box, refer to the info section for the tree builder
* move the clitic instructions from the "inputs" info to the tree builder info
* if possible, also auto-click on the tree builder info to expand it when a hyphen is found and the pop up is triggered

PR #330, opened 2020-03-10
Add exhaustParent to the interface. Move exhaustChild to show more section.

ColinAChen:
I didn't change the description of exhaustParent.


interface1.html
```diff
@@ -1312,6 +1312,16 @@ <h3>...counting leaves</h3>
 					</div>
 
 					<div onclick="showMore('moreLayeringConstraints')" class="show-more">Show more...</div>
+					<div id="moreLayeringConstraints" class="more-constraints">
```
jbellik: This constraint should've been added to the existing "show more" section below. As it is, the remaining "show more" constraints just don't seem to be available.

PR #332, opened 2020-03-10
first pass on match any constituent. Added two options for match. Use…

### LINKED ISSUE #321: Custom Match: any constituent
What if we also had custom Match(α,β) where α and/or β could be "any constituent"? For example, Match(XP,constituent) would be fine with:

[a b] → [.w a b]
[a b] → (.phi a b)
[a b] → {.i a b}

And also fine with:

[a] → a

I suppose the same could be done for Align and Wrap.

ColinAChen:
… options to determine whether or not to check that the syntactic and prosodic categories need to be matched
resolve #321

### DIFF
--- a/constraints/match.js
+++ b/constraints/match.js
@@ -126,7 +126,7 @@ function matchSP(inputTree, pTree, sCat, options)
 	*  - either it is lexical (sParent.func is false) OR requireLexical is false
 	*  - either it has an overt head (sParent.silent is false) OR requireOvertHead is false
 	*/
-	if(((sParent.cat === sCat ) && !options.anySyntacticConstituent)
+	if((options.anySyntacticConstituent || (sParent.cat === sCat ))
 	&& !(options.requireLexical && sParent.func)
 	&& !(options.requireOvertHead && sParent.silentHead)
 	&& !(options.maxSyntax && !sParent.isMax)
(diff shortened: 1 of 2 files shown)


constraints/match.js
```diff
@@ -124,7 +126,7 @@ function matchSP(inputTree, pTree, sCat, options)
 	*  - either it is lexical (sParent.func is false) OR requireLexical is false
 	*  - either it has an overt head (sParent.silent is false) OR requireOvertHead is false
 	*/
-	if(sParent.cat === sCat
+	if(((sParent.cat === sCat ) && !options.anySyntacticConstituent)
```
jbellik: I think the logic here actually needs to be:
if(((sParent.cat === sCat ) || options.anySyntacticConstituent) ...
because if we're satisfied with any syntactic category, then it's also not a problem if it is some specified category. 
Another thing to watch out for: what if anySyntacticConstituent === true and then sCat isn't defined? 
maybe best to do the opposite order of conditions to handle that: 
if((options.anySyntacticConstituent || (sParent.cat === sCat )) ...

constraints/match.js
```diff
@@ -161,7 +163,7 @@ function hasMatch(sNode, pTree, options)
 
 	var sLeaves = getLeaves(sNode);
 	markMinMax(pTree);
-	if(catsMatch(sNode.cat, pTree.cat)
+	if((catsMatch(sNode.cat, pTree.cat) && !options.everyProsodicConstituent)
```
jbellik: This should also be an OR. See note above.

PR #340, opened 2020-03-16
Input gen options

### LINKED ISSUE #247: Automatic input generation function
It would be great if we could automate input generation, in a way similar to GEN. For two integers _min_ and _max_, get all of the strings of words [1, 2, ..., _min_], [1, 2, ..., _min_+1], ..., [1, 2, ..., _max_], labeling the terminals alphabetically ([a], [a b], [a b c], [a b c d], etc.).

Then, given those strings, generate all of the possible input trees, ideally with two limits on branching that the user could select: _i_-ary through _j_-ary. If arbitrary _i_ and _j_ would be too hard, it would be great if we could have _i_ = 1 or _i_ = 2, and _j_ = 2.

Like with GEN (but syntactified), we would want a root node cp, non-root non-terminals xp's, and terminal w's. And like with prosodic trees, we wouldn't want any unary-branching xp dominating another xp. The only possible unary branching xp should be of the form [w].

Finally, it would be good to have the option of including or excluding the root cp (i.e. replacing it with xp). I don't think we would be interested in considering inputs with **and** without a cp-root at the same time, just sets of inputs that all have cp-roots or that all have xp-roots.

ivpetkov:
I implemented everything from the most recent mockup except I can't figure out how to make the text in the terminal string boxes persist when you add a new one.
Closes #247


jbellik:
I think you might have to do something similar to what the tree builder does when it wants to add something to the inner html. Take a look at the functions in lines 68-110 (toInnerHtmlFrags and following).

jbellik:
"allow adjuncts" is backwards: 
* when it's checked, "a b" should yield [[a]b], [a [b]], [[a][b]] --> this more permissive setting should be default
* when not checked, "a b" should yield [[a] b], [a [b]] --> not the default

jbellik:
* Show trees by default when you click generate trees (along with displaying the "up to date" message)
* Move "syntax parameters" to the top and put "generate trees" next to "add terminal string"
* Consider making "generate trees" orange

ivpetkov:
I accidentally merged the mirror-images branch into this branch when I was trying to merge the other way around.

jbellik:
Ok, so this one isn't currently waiting for a review update, is it?

ivpetkov:
Yes, I'll fix the update message.

Issue #312, opened 2020-02-24
Input Gen option: Remove mirror images


nkalivoda:
Input Gen option to remove trees that are mirror images of trees earlier in the list.

Definition of mirror image:
Ignoring ids, but not cats, the mirror image of a tree T is a tree T' in which every children array in T is in reverse order in T'.

Example:
T  = [x0 [x0 [x0]]]
T' = [[[x0] x0] x0]

Reason to have this option:
If Con doesn't contain any mirror-distinguishing constraints (e.g. AlignL, AlignR, StrongStart), then for every language in the typology, a mapping sTree→pTree entails the mapping mirror(sTree)→mirror(pTree). Unnecessary to look at all of the mirrors (which can be numerous).

Not equivalent to imposing a left- or right-headedness requirement. If "remove mirrors" were selected, then one of either [x0 [[x0] x0]] or [[x0 [x0]] x0] would be an input (whichever one happens to be earlier in the unfiltered list), even though neither of these is generated when headedness is restricted to one side.

nkalivoda:
This should only apply if sTree≠mirror(sTree). We wouldn't want to exclude a tree like [[x0] [x0]].

nkalivoda:
Question that arose in meeting: Is it possible to write a phrase structure grammar that doesn't generate mirror(T) when T≠mirror(T)?

jbellik:
Currently, this seems to remove symmetrical outputs -- it is pruning out too many.
Allow mirrors: [[a] b], [a [b]], and symmetric [[a][b]]
Exclude mirrors: should yield [[a] b] and symmetric [[a][b]] (excluding [a [b]] -- or we could keep this one and exclude [[a] b])
Currently exclude mirrors only yields: [[a] b] and incorrectly leaves out [[a][b]]

jbellik:
For the three word case, too many are excluded as well. Checking "remove mirror images"
should include the first six (highlighted in blue), not just the first three as it currently does.
![image](https://user-images.githubusercontent.com/12523306/78182529-35933400-741b-11ea-8f7e-4e93c0c05683.png)

PR #342, opened 2020-04-08
Merge mirror-images into input-gen-options

ivpetkov:
I think the "remove mirror images" option is working (issue #312)

I don't know if it's better to merge mirror-images into input-gen-options then merge input-gen-options into master or update mirror-images from input-gen-options and merge mirror-images to master.

### DIFF
--- a/syntaxFilters.js
+++ b/syntaxFilters.js
@@ -73,13 +73,23 @@ function mirrorImages(sTree, sTreeList) {
 	var reverseTree = JSON.parse(JSON.stringify(sTree));
 	// console.log(sTree)
 	// console.log(parenthesizeTree(sTree))
-	reverseTree = getReverseTree(reverseTree);
+	// reverseTree = getReverseTree(reverseTree);
 	// console.log(parenthesizeTree(sTree))
+	console.log("start", index)
+	console.log("original rTree")
+	console.log(reverseTree)
 	for(var i = 0; i < index; i++) {
 		var currTree = sTreeList[i];
-		var checkReverse = checkReverseTree(reverseTree, currTree);
-		console.log(checkReverse, "reverseTree: ", index, "currTree: ", i)
-		if(checkReverse) {
+		console.log("original sTree")
+		console.log(currTree)
+		console.log("--------------------------")
+		// var checkReverse1 = checkReverseTree(reverseTree, currTree);
+		var checkReverse1 = testing(currTree, reverseTree);
+		console.log(checkReverse1)
+		// var checkReverse2 = checkReverseTree(currTree, reverseTree);
+		// console.log(checkReverse2)
+		// console.log(checkReverse, "reverseTree: ", index, "currTree: ", i)
+		if(checkReverse1) {
 			mirrorImageExists = true;
 			return mirrorImageExists;
 		}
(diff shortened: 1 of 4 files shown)


jbellik:
Oops I didn't read these comments before merging input gen options into master!

jbellik:
see test/mirror_test.hml for a demonstration of this problem

checkMirror gives false positives under some conditions that I haven't yet identified.

ivpetkov:
Thanks for making a test file! I probably should've done that earlier.

ivpetkov:
The problem was the way checkMirror() was traversing some trees, so now there is another variable isMirror that checks if at any point in checkMirror() the 2 categories weren't equal. checkMirror() still sometimes returns false positives in the test file, but when you make the same comparison using mirrorImages() it returns correctly because isMirror is checked in mirrorImages().

syntaxFilters.js
```diff
@@ -64,65 +64,45 @@ function headsOnWrongSide(sTree, side, strict){
...
-			var rCat = rChild.cat;
-			var currCat = currTree.children[length-1-i].cat;
-			if(rCat && currCat) {
-				if(rCat === currCat) {
-					checkTree = true;
-				}
+// check for if trees are mirror images of eachother
+function checkMirror(sTree, rTree) {
+	var equal = true;
+	if(sTree.children && sTree.children.length){
```
jbellik, line 90: If sTree has no children, this will always return true. :-/ We need to evaluate to false if sTree has no children but rTree does.

syntaxFilters.js
```diff
@@ -64,65 +64,45 @@ function headsOnWrongSide(sTree, side, strict){
...
-// returns true if tree is a mirror of an eariler tree
-// returns false if tree is not a mirror image of an earlier tree
+var isMirror = true;
+// returns true if tree is a mirror of an eariler tree in list
+// returns false if tree is not a mirror image of an earlier tree in list
 function mirrorImages(sTree, sTreeList) {
+	isMirror = true;
 	var mirrorImageExists = false;
 	var index = sTreeList.indexOf(sTree);
 	var reverseTree = JSON.parse(JSON.stringify(sTree));
```
jbellik, line 74: 1. This variable name confuses me every time. The tree hasn't been reversed, has it?
2. Why stringify + parse rather than using the tree directly?

syntaxFilters.js
```diff
@@ -64,65 +64,45 @@ function headsOnWrongSide(sTree, side, strict){
 
 }
 
-// Input Gen option to remove trees that are mirror images of trees earlier in the list
-// returns true if tree is a mirror of an eariler tree
-// returns false if tree is not a mirror image of an earlier tree
+var isMirror = true;
```
jbellik, line 67: I'm worried about strange scope issues arising from declaring var isMirror out here. I think it would be better to declare it inside mirrorImages(), pass it to checkMirror(), and return it to mirrorImages().

syntaxFilters.js
ivpetkov: 1. Yes, it hasn't been reversed, sorry about that. The reverseTree variable name is left over from when I had a function that reversed the tree first and then checked to see if it was equal. I just forgot to change it after I changed the approach.
2. This is also left over from when I was reversing the tree before comparing trees. I ran into a issue where the original tree was being modified in the reverse function, so this was my way of copying the tree. 

Sorry about these, I think I just got used to the code being this way that I didn't realize it wasn't necessary anymore.

syntaxFilters.js
ivpetkov: I agree, this was an admittedly hacky fix.

ivpetkov:
Everything looks good to me! This solution is a lot cleaner and better than what I had before.

jbellik:
Ok, great, thanks for looking through it! I had @ozanblk  look over it and he figured out the major bug which was that later children in a set could overwrite previous "false" values. After that it was easy to clean things up.

PR #344, opened 2020-04-11
Wrap PS

### LINKED ISSUE #307: Wrap PS constraint
Create a prosody --> syntax version of wrap:
wrapPS(stree, ptree, cat) = Assign a violation for every node P in ptree that is of category cat for which there is no node S in the syntactic tree with category categoryMapping[cat] that contains all the terminals dominated by P.

On the interface, this should go directly under AlignPS.

ColinAChen:
WrapPS implemented, added to interface, and added to test file. resolve #307

### DIFF
--- a/interface1.html
+++ b/interface1.html
@@ -974,10 +974,11 @@
 								</div>
 							</div>
 							<div class="category-row">
-								<div class="category-selection-div"><input type="checkbox" name="category-wrap" value="cp">CP</div>
-								<div class="category-selection-div"><input type="checkbox" name="category-wrap" value="xp" checked="checked">XP</div>
-								<div class="category-selection-div"><input type="checkbox" name="category-wrap" value="x0">X<sup>0</sup></div>
+								<div class="category-selection-div"><input type="checkbox" name="category-wrap" value="i">&iota;</div>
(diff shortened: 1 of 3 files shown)


interface1.html
```diff
@@ -966,6 +966,19 @@ <h3>Align/Wrap Theory</h3>
...
 							</div>
+							<div class="constraint-selection-table">
+							<div class="constraint-row">
+								<span><input type="checkbox" name="constraints" value="wrapPS">Wrap(Prosody&rarr;Syntax)</span>
+								<div class="info">
+									<span class="content">Assign one violation for every node S of category K in the syntactic tree that does not have a corresponding node P in the prosodic tree, where P is of the category K corresponds to and P contains all the terminals dominated by S. SPOT function: wrap(). (Truckenbrodt 1995, 1999)</span>
+								</div>
+							</div>
+							<div class="category-row">
+								<div class="category-selection-div"><input type="checkbox" name="category-wrap" value="cp">CP</div>
```
jbellik, line 977: These categories need to be prosodic categories: i, phi, w.

jbellik:
In general, when you run a test and it gives you back 0s in every row, that should be an indication that something is wrong. Here, you called wrapPS on a syntactic category. It needs to operate on prosodic categories.

PR #355, opened 2020-04-24
Align focus

### LINKED ISSUE #318: Align-Focus
Truckenbrodt (1999) has a constraint Align-Foc. His exact words:

Screen Shot 2020-02-25 at 10 22 52 AM

There is also a left-aligning version.

To do this, we should have a "foc" attribute, which can be freely placed on terminal and non-terminal nodes of the sTree.

The constraints should have "foc" as the "sCat" (technically an "sAttribute"), and ideally would have a custom pCat argument. Something like:

**Align(sTree,pTree,foc,pCat,direction)**
For every node S with attribute "foc" in sTree, assign a violation if the _direction_-edge of S is not aligned to the _direction_-edge of a node of category pCat in pTree.

ColinAChen:
close #318

### DIFF
--- a/constraints/alignWrap.js
+++ b/constraints/alignWrap.js
@@ -20,6 +20,7 @@ function alignRight(sTree, pTree, sCat){
 function alignSP(sTree, pTree, sCat, d){
 	var getEdge = (d==="left") ? getLeftEdge : getRightEdge;
 	var vCount = 0;
+	
 	walkTree(sTree, function(sNode){
 		if(sNode.cat !== sCat)	 // only go further if sNode has the category we're interested in
 			return;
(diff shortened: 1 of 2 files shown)


constraints/alignWrap.js
```diff
@@ -66,6 +67,35 @@ function alignRightPS(sTree, pTree, cat){
...
+		if(!sNode.foc)	 // only go further if sNode is a focus node
+			return;
+		var sEdge = getEdge(sNode);
+		if(!sEdge)
+			sEdge = sNode;	// If sNode is a leaf (which it probably shouldn't be but depending on the tree might be),
+								// then look for a p-node that matches sNode itself. TODO is this a good idea?
+		var noMatch = true;
+		walkTree(pTree, function(pNode){
+			//!catsMatch(sCat, pNode.cat)
+			if(pNode.cat === cat)
```
jbellik, line 83: This should be a return if the categories !==

PR #356, opened 2020-04-28
Gradient noShift

### LINKED ISSUE #293: Gradient NoShift as Linearity
If we try to implement a gradient NoShift, one possibility would be to use McCarthy & Prince's (1995) constraint Linearity.

> **Linearity** (McCarthy & Prince 1995, "Faithfulness and Reduplicative Identity")
> S₁ is consistent with the precedence structure of S₂, and vice versa.
> Let x, y ∈ S₁ and x', y' ∈ S₂.
> If xℜx' and yℜy', then
> x < y iff ¬(y' < x').

S₁ is the input string, S₂ is the output string, 'ℜ' is the correspondence relation, and 'p<q' means 'p precedes q'.

SPOT version:

> **Linearity** (McCarthy & Prince 1995 applied to terminal strings)
> For every pair of terminals (x,y), assign a violation if x precedes y in the terminal string of sTree and y precedes x in the terminal string of pTree.

Example with moving X:
Screen Shot 2020-02-08 at 12 00 39 PM

Selected candidates with potential movers X,Y:
Screen Shot 2020-02-08 at 12 00 47 PM

Please correct me if I have misunderstood or miscounted anything.

ivpetkov:
I tested it in noShiftGradientTest.html with the examples that Nick gave in the issue. The function in that file is the same as the noShiftGradient function in noShift.js, I just modified it to take in the lists of terminals so I didn't have to make all the trees to test it. What other cases should I test? Issue #293

### DIFF
--- a/constraints/noShift.js
+++ b/constraints/noShift.js
@@ -1,26 +1,31 @@
 function noShift(stree, ptree, cat){
     //Get lists of terminals
 
-    var sleaves = getLeaves(stree);
-    var pleaves = getLeaves(ptree);
-    var sorder = new Array(sleaves.length);
-    var porder = new Array(pleaves.length);
+    // var sleaves = getLeaves(stree);
+    // var pleaves = getLeaves(ptree);
+    // var sorder = new Array(sleaves.length);
+    // var porder = new Array(pleaves.length);
+    //
+    // try {
+    //   if(sleaves.length != pleaves.length) {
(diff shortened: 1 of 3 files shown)


ivpetkov:
For adding linearity to the interface, a made a branch off of this branch. Pull request #361 should be merged before this pull request because it merges linearity-interface into gradient-noshift.

Issue #249, opened 2020-01-21
GEN option: restrict trees to being n-ary (user specifies n)


nkalivoda:
Could we add a GEN option to limit prosodic branching to be maximally binary? (Like "Only generate branching constituents", but for BinMax(pCat,branches) instead of BinMin.)

jbellik:
Add a GEN option similar to noUnary that defines an upper limit on the number of branches a node can have (i.e., if you set it to k, you will get trees that are maximally k-ary).

jbellik:
Simple implementation: use the function ternaryNodes() in syntaxFilters.js to exclude candidates in window.GEN_impl that have a node with more than maxBranches children.

PR #362, opened 2020-05-05
maxBranching GEN option

ivpetkov:
Issue #249. In the test file maxBranchingTest.html I have trees {a b} and {a b c}. When the variable max_branches is 2, it excludes some candidates from {a b c} and none from {a b}. The excluded candidates are written to the console.


ivpetkov:
I added interface support (issue #366)

Issue #367, opened 2020-05-06
input GEN clitic categories


jbellik:
If generating syntactic trees, set category of nodes suffixed "-clitic" to "clitic" instead of "syll"

PR #369, opened 2020-05-12
syntax sensitivity for clitics

maxTarlov:
fix for issue #367


maxTarlov:
Fixed an issue with hyphens in built-in analyses. It might be possible to have this error ignore built-in analyses, but it was a lot easier to just remove the hypthens

PR #370, opened 2020-05-12
separated syntax-prosody mapping constraints in fieldsets

ivpetkov:
Issue #365. I just reorganized the existing constraints. Should I add "show more/less" for the "match" or "align/wrap" fieldsets?


jbellik:
Yes, please. Everything that was in the "show more/less" section previously should go in a "show more/less" section now too. Thanks!

PR #378, opened 2020-05-20
merge 'error-handling' changes into 'new-error-handling'

ivpetkov:
'new-error-handling' is just a copy of master


ivpetkov:
Removed changes to candidategenerator.js because conflicts were too complicated

Issue #373, opened 2020-05-14
noAdjacentHeads options on interface


jbellik:
Currently, with auto input generation on the interface, noAdjacentHeads is always true; there is no way to set it to false. This creates two problems:

1. If you want to build compound word syntaxes by picking x0 as the intermediate category, you get no results (since this would require adjacent heads)

2. We want to add an option to leave out unary XPs and turn them into plain x0s, which would also require adjacent heads in the outputs. [[a] b] --> [a b]

jbellik:
Let's change interface1.js to automatically set noAdjacentHeads to false if you pick either of the following:
- intermediate category : x0
- unary XPs are invisible (new option that we need to implement, which sets noUnary:true, noAdjacentHeads: false)

maxTarlov:
do we want the check to happen when the value of intermediateCategory/terminalCategory/noUnary changes or when sTreeGEN is run? If we run this visually in the interface (on change) I can disable the noUnary option while while the other options are incompatible.

maxTarlov:
correction: I would set the noAdjacentHeads input to disabled, not noUnary.

PR #380, opened 2020-05-26
Adjacent heads

maxTarlov:
Addresses #373 . When intermediate syntactic category, terminal syntactic category or noUnary is selected, syntacticOpCompatible() checks that all the options are compatible. If not, it sets adjacentHeads to true and disables the input. If the options were previously incompatible and now aren't, syntacticOpCompatible re-enables adjacentHead and sets it back to what it was before the options were incompatible.
My thinking behind this is that someone might select noUnary, not see that it had selected adjacentHeads, and then deselect noUnary. In that case, the user would get adjacent heads where they would have expected not to.

### DIFF
--- a/interface1.js
+++ b/interface1.js
@@ -1062,12 +1062,12 @@ function changeInputTabs(from, to) {
 }
 
 function syntaxOpCompatible(){
-	let spotForm = document.getElementById("spotForm");
-	let recursive = spotForm["autoInputOptions-recursiveCategory"].value;
-	let terminal = spotForm["autoInputOptions-terminalCategory"].value;
-	let unary = document.getElementById("unarySyntaxOption").checked;
+	spotForm = document.getElementById("spotForm");
+	recursive = spotForm["autoInputOptions-recursiveCategory"].value;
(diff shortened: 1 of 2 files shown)


maxTarlov:
I think "Unary XPs are invisible" should be changed to "Unary intermediate nodes are invisible"

maxTarlov:
Removed noAdjacentHeads option from interface and moved desired behavior to button onClick function autoInputTree() in interface.js

interface1.js
```diff
@@ -714,6 +714,10 @@ window.addEventListener('load', function(){
 			autoInputOptions.recursiveCategory = spotForm['autoInputOptions-recursiveCategory'].value;
 			autoInputOptions.terminalCategory = spotForm['autoInputOptions-terminalCategory'].value;
 
+			if(autoInputOptions.recursiveCategory == autoInputOptions.terminalCategory || autoInputOptions.noUnary){
```
jbellik, line 717: === and change this to check specifically for recursiveCategory==='x0'

PR #405, opened 2020-06-25
split on comma, remove whitespace, add all elements after first as no…

### LINKED ISSUE #399: Accent in interface
Some constraints (accentAsHead, noLapseL), as well as the tone annotation, rely on an accent attribute on words (nodes where cat: 'w'). Currently this attribute can only be added on the interface by going into the code view and adding it directly to the JS code. 

Provide the same functionality for accent as for silentHead and func. That is, if the cat field for a node in tree builder is labeled "x0, accent", then in the code, that node should be represented as:

{cat: 'x0', accent: true} (+ whatever children and id it has)

image

ColinAChen:
…de attributes

#resolve #399

### DIFF
--- a/candidategenerator_wrapper.js
+++ b/candidategenerator_wrapper.js
@@ -96,7 +96,8 @@ window.GEN = function(sTree, words, options){
 				}
 				var accentSuffix = '';
 				if(words[i].accent){
-					accentSuffix = '-'+words[i].accent;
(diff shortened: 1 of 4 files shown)


ColinAChen:
If an input is invalid should it be sanitized (remove invalid characters) then turned into an attribute or ignored

jbellik:
It would be best to sanitize it, I think.

PR #407, opened 2020-06-30
Bar levels option

### LINKED ISSUE #390: Input GEN option: X'
Add an option on the interface called:

"Bar levels (X', C') are invisible" 

Info: "By default, SPOT treats bar level projections the same as phrase level projections, and labels them all as XP (or CP). This makes them visible to Match and Align constraints. On this setting, all syntactic trees are binary branching. To treat bar level projections as invisible to mapping constraints, check this box, which will allow GEN to also create ternary branching syntactic trees, in which X' (or C') are not represented at all. An example of a tree in which the X' level is not represented at all: [[spec] head [comp]]. Under default settings, this tree would instead be represented as: [[spec][head [comp]]]."

In trying to decide where to put this option on the interface, I realized we should reorder things a little. I'd like the Syntax Parameters go be ordered as follows:

- Root syntactic tree in...
- Intermediate nodes are...
- Syntactic terminals are...
- Allow adjuncts
- Bar levels are invisible
- Unary XPs are invisible
- Head requirements
- Remove mirror images

The effect of selecting "Bar levels are invisible" on the backend should be to set maxBranching = 3 (instead of the default value of 2).

This option should only be available if the intermediate category is xp or cp, not if the intermediate level is x0. If x0 is selected as the intermediate category, then "Bar levels are invisible" should be greyed out and disabled.

In addition, selecting "Bar levels are invisible" should change the options available under "Head requirements." When "Treat bar levels as invisible" is *not* selected, then we should disable "Heads must be on the left edge" and "Heads must be on the right edge". When "Treat bar levels as invisible" *is* selected, then those options should be enabled.

ivpetkov:
closes #390

### DIFF
--- a/inputCandidateGenerator.js
+++ b/inputCandidateGenerator.js
@@ -42,6 +42,9 @@ function sTreeGEN(terminalString, options)
     var sTreeList = autoSTreePairs.map(x=>x[1]);
 
     //Apply filters
+    if(options.noBarLevels){
+      options.maxBranching = 3;
+    }
     if(options.allowClitic){
       var cliticTrees = getCliticTrees(terminalString, options);
       if(cliticTrees) {
--- a/inputCandidateGenerator.js
+++ b/inputCandidateGenerator.js
@@ -73,7 +73,9 @@ function sTreeGEN(terminalString, options)
     if(options.noMirrorImages){
       sTreeList = sTreeList.filter(x => !mirrorImages(x, sTreeList));
     }
-
+    if(options.noBarLevels){
+      sTreeList = sTreeList.filter(x => !threeXPs(x));
+    }
     return sTreeList;
 }
 
--- a/inputCandidateGenerator.js
+++ b/inputCandidateGenerator.js
@@ -57,7 +57,7 @@ function sTreeGEN(terminalString, options)
         sTreeList = sTreeList.filter(x => !x0Sisters(x, 'x0'));
     }
     if(options.noAdjuncts){
-        sTreeList = sTreeList.filter(x => !x0Sisters(x, options.recursiveCategory));
+        sTreeList = sTreeList.filter(x => !containsAdjunct(x));
     }
     if(options.maxBranching > 0){
         sTreeList = sTreeList.filter(x=>!ternaryNodes(x, options.maxBranching));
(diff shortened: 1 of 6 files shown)


interface1.js
```diff
@@ -722,7 +722,7 @@ window.addEventListener('load', function(){
 			if(autoInputOptions.recursiveCategory === 'x0' || autoInputOptions.noUnary){
 				autoInputOptions.noAdjacentHeads = false;
 			}
```
jbellik, line 724: As far as I can tell, you're not adding the information from noBarLevels to autoInputOptions here.

inputCandidateGenerator.js
```diff
@@ -42,6 +42,9 @@ function sTreeGEN(terminalString, options)
     var sTreeList = autoSTreePairs.map(x=>x[1]);
 
     //Apply filters
+    if(options.noBarLevels){
+      options.maxBranching = 3;
```
jbellik, line 46: I was thinking we would set this in interface1.js, rather than adding noBarLevels as an option to sTreeGEN() itself.

jbellik:
Additional filter to be implemented in syntaxfilters.js:

Return true if there is any node that has more than two children x such that x.cat === 'xp'. Two xp children is fine, but three (or more) is not fine.

Ex: In the examples below, [ ] stand for the boundaries of a node where cat === 'xp'
threeXPs([[a][b][c]] ) --> true
threeXPs( [[a][b]] ) --> false
threeXPs( [a [b] [c] ] ) --> false

When noBarLevels == true, then: maxBranching = 3, excludeThreeXPs = true
Add a filter to sTreeGEN that runs threeXPs and excludes trees that threeXPs evaluates to true.

ivpetkov:
I changed it so maxBranching is set in interface1.js. The problem with how I had it in sTreeGEN was that it set maxBranching after GEN was called.

jbellik:
Ok apparently I can only comment on lines you added. Line 60 in inputCandidateGenerator.js is where you'll need to put in the new filter I'm asking you to write. Write it in syntaxFilters.js. You can call it containsAdjunct. It should return true if there is a node in it whose children are all xps, false if all nodes have an x0 child.

jbellik:
containAdjunct: Return true if there is an xp whose children are all xps. Return false otherwise.

Issue #404, opened 2020-06-24
Add interface support for marking left-aligned prosodic heads


jbellik:
In "Tree display options" under "Mark heads", add options to implement interface support for the functionality in #403 

This should be radio buttons for left-aligned heads vs. right-aligned heads (default value). Like the options for tone annotation, they should be hidden unless "Mark heads" is selected.

maxTarlov:
Closed with PR #409

PR #409, opened 2020-06-30
Head marking

### LINKED ISSUE #403: Head marking: head alignment option
The feature implemented in #275 is appropriate for Japanese, and always puts the head on the right. Other languages have left-aligned heads. 

Extend the head marking function to take a parameter headSide = left or right. Its default value should be right. If headSide = right, use the existing head marking function. If headSide = left, then the head is the leftmost node of the highest pcat in that set of sisters, instead of the rightmost.

maxTarlov:
Closes #403 and #404


maxTarlov:
As it stands binMaxHead assumes right-headedness. Presumably it should also be able to work with left-headedness, but should the user supply that information manually or can we scrape it from the handedness of the head marking chosen in the interface?

jbellik:
Max will try to resolve this conflict and if that doesn't work, Jenny will then try.

maxTarlov:
Merge conflicts resolved, ready to merge automatically

PR #419, opened 2020-07-21
Major minor phrases

### LINKED ISSUE #263: GEN and CON for MiPs and MaPs
New version of GEN with Minor Phrases (MiPs) and Major Phrases (MaPs).

No recursion possible. PH = [ι > MaP > MiP > ω]. Correpsondences: cp ↔ ι, xp ↔ MaP, xp ↔ MiP, x0 ↔ ω.

If Exhaustivity is on, then every ω must be in a MiP. If not, then every ω can be sister to MiP (= daughter of MaP) or sister to MaP (= daughter of ι).

All constraints that refer to φ should be able to refer to φ = MiP/MaP, just MiP, or just MaP.

ColinAChen:
close #263

### DIFF
--- a/prosodicHierarchy.js
+++ b/prosodicHierarchy.js
@@ -12,10 +12,27 @@ var categoryPairings = {
 function catsMatch(aCat, bCat){
 	if(aCat === undefined && bCat === undefined)
 		return true;	//vacuously true if both nodes are leaves (with no category)
-	else if(categoryPairings.hasOwnProperty(aCat))
-		return categoryPairings[aCat] === bCat;
-	else if(categoryPairings.hasOwnProperty(bCat))
-		return categoryPairings[bCat] === aCat;
+	// aCat is the key
+	else if(categoryPairings.hasOwnProperty(aCat)){
+		// check if bCat is an array
+		if (Array.isArray(bCat)){
+			return bCat.includes(aCat);
+		}
+		else{
+			return categoryPairings[aCat] === bCat;
+		}
+	}
+	// bCat is the key
+	else if(categoryPairings.hasOwnProperty(bCat)){
+		// check if aCat is an array
+		if (Array.isArray(aCat)){
+			return aCat.includes(bCat);
+		}
+		else{
+			return categoryPairings[bCat] === aCat;
+	
+		}
+	}
 	else
 	{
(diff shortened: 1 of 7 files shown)


prosodicHierarchy.js
```diff
@@ -29,6 +58,17 @@ var sCat = ["cp", "xp", "x0"];
 //defines the prosodic hierarchy
 var pCat = ["u", "i", "phi", "w", "Ft", "syll"];
 
+//defines Major Minor Phrase prosodic hierarchy
+var MaPMiPpCat = ["i", "MaP", "MiP", "w"];
+
+function setPCat(newPCat){
+	pCat = newPCat;
```
jbellik, line 65: We need to add the private functions pCat.isHigher and pCat.isLower to this too, I think.

prosodicHierarchy.js
```diff
@@ -29,6 +58,17 @@ var sCat = ["cp", "xp", "x0"];
...
+}
+
+function resetPCat(){
+	pCat = ["u", "i", "phi", "w", "Ft", "syll"];
+
```
jbellik, line 70: here also

test/gen_cat_test.html
```diff
@@ -26,18 +26,27 @@ <h2>GENerating inputs</h2>
 	var MatchSPMatchPS = ['binMaxBranches-phi', 'binMinBranches-phi', 'matchSP-xp'];
 	
 	/*A list of names for constraint sets.*/
-    var conNames = ['MatchSPMatchPS'];
-	
+	    var conNames = ['MatchSPMatchPS'];
+	var customPCatList = ['i', 'MaP', 'MiP', 'w'];
+	var customPairings = {
+		"clause": "i",
```
jbellik, line 32: customPairings doesn't match customPCatList --> we expect errors

ColinAChen:
image
I think the code for this warning may need to be changed to look at the new PH_PHI constant pCat because pCat may get overwritten with a custom pCat.

jbellik:
Yes, and it should really say "[category] is not in the currently specified prosodic hierarchy. See PH_PHI and PHI_MAJMIN in prosodicHierarchy.js for some possible predefined prosodic hierarchies."

Ideally this error would actually also display the current pCat.

jbellik:
Just tried this out and I think we need that warning "category argument phi is not in pCat" to be an error, actually. Right now it's displaying once for every candidate, I believe, and instead we should be quitting as soon as the first mismatch is discovered. That error should also provide some more information by displaying the current pCat.

PR #424, opened 2020-08-04
Terminal strings interface

### LINKED ISSUE #402: Terminal strings on interface
Add the functionality for Issue #401 to the interface in "GEN: Input settings" to automatically generate a set of terminal strings. These terminal strings will then be the inputs to streeGEN.

ivpetkov:
Closes #402

### DIFF
--- a/interface1.js
+++ b/interface1.js
@@ -919,6 +919,7 @@ window.addEventListener('load', function(){
 		return htmlChunks.join('');
 	}
 
+	// add thicker line between generated strings of different lengths
 	function addThickLine(genStringsList) {
 		var sheet = document.styleSheets[1];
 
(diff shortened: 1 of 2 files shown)


interface1.js
```diff
@@ -798,6 +804,155 @@ window.addEventListener('load', function(){
...
+		genTerminalStrings();
+		document.getElementById('genStringsArea').style.display = 'block';
+		document.getElementById('gen-strings-switch').checked = true;
+		document.getElementById('strings-switch-text').innerHTML = 'Hide generated terminals strings';
+	});
+
+	var genStringsList;
+
+	// generate and display terminal strings
+	function genTerminalStrings() {
```
jbellik, line 846: This is minor but I'm just noticing that we have genTerminalStrings() (a UI function) plus generateTerminalStrings() (a purely backend function that genTerminalStrings() calls). This is a recipe for confusion. I think it would be good to rename genTerminalStrings() to something like genTerminalStringsUI() or genTerminalStringsInterface().

PR #427, opened 2020-08-05
Syntax parameters

### LINKED ISSUE #423: "Syntax parameters"
A. Make this fieldset expanded by default

B. Break it into 3 parts:
1. Syntactic parameters: Root syntactic tree in, Intermediate nodes are, Terminals are, Include adjunction (currently "Allow adjuncts"), Head requirements
2. Clitics: Add XP clitics directly under the root, Label some terminals as clitics (currently "Allow clitic in terminal string")
3. Visibility to phonology: Treat unary XPs as X0s (currently "Unary XPs are invisible"), Treat intermediate projections (bar levels) as phrasal (currently "Bar levels are invisible"; need to flip the logic on this from the current configuration -- make it checked by default and flip the true value to false and vice versa).

Currently not sure where to put the "Remove mirrors" option.

More refinements to come...

ColinAChen:
close #423

### DIFF
--- a/inputCandidateGenerator.js
+++ b/inputCandidateGenerator.js
@@ -14,6 +14,8 @@
 *    the resulting syntactic trees. Default = 2
 *  - minBranching: determines the maximum number of branches that are tolerated in
 *    the resulting syntactic trees. Default = 2
+*  - noBarLevels: if false (default), bar levels are treated as phrasal. 
+*    If true, bar levels are not represented, and ternary branching is permitted.
 *  - addClitics: 'right' or 'left' determines whether clitics are added on the
 *    righthand-side or the left; true will default to right. false doesn't add any clitics.
 *    Default false.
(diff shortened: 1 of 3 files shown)


ColinAChen:
If I am understanding this bit of code correctly, the loop is used to find the index of the noBarLevels option from the list of autoIntputOptions. Is there a reason we are not using the indexOf method to achieve this?

image

ColinAChen:
Please disregard my last comment, I understand why we need to use a loop to find the noBarLevels option. All the logic using noBarLevels should be reversed.

inputCandidateGenerator.js
```diff
@@ -73,7 +73,7 @@ function sTreeGEN(terminalString, options)
     if(options.noMirrorImages){
       sTreeList = sTreeList.filter(x => !mirrorImages(x, sTreeList));
     }
-    if(options.noBarLevels){
+    if(!options.noBarLevels){
```
jbellik, line 76: This works, but it would be better to leave the logic in sTreeGEN() the same, and flip the value of noBarLevels before passing it in as an option. I was intending this change to be entirely on the interface. I just want to be sure that if nothing is specified for noBarLevels, sTreeGEN() creates binary branching trees rather than ternary branching trees.

PR #428, opened 2020-08-05
auto input save/load

maxTarlov:
saving and loading work for auto input gen, but clearing doesn't yet


maxTarlov:
clearing also works now

built-in_analyses.js
```diff
@@ -1,5 +1,42 @@
 /* Built-in Analyses */
 
+function clearInputs(){
+  let inputOptions = spotForm['autoInputOptions'];
```
jbellik, line 4: FYI "let" doesn't work on IE, but we don't support that browser anyway.

jbellik:
I will go ahead and merge this because of the other pull requests that include its changes, and because buggy saving/loading of the auto input gen is better than no saving/loading. But there are still some bugs to work out.

PR #446, opened 2020-08-26
Save/load bugs

maxTarlov:


maxTarlov:
The bug you noticed was a really simple fix, but I also noticed another issue with the output GEN option "Restrict maximum number of branches" which takes a text input in a way I did not expect. That is all fixed now too. I am starting to think we need a more robust way to test this particular feature since it is so fragile, can we add this to the agenda for Wednesday's meeting?

jbellik (approved):

PR #447, opened 2020-08-29
CTConstraints

3ddiehead:
Added 4 functions to ctConstraints.js:
- notMutualCommand(tree, x, y) // takes a tree and two nodes and returns true if nodes are not mutually commanding
- dominates(node, x) // takes two nodes and returns true if x.id is in the descendants of 'node'
- pairExists(pairs, x, y) // takes an array of arrays as well as two items and returns true if there are any arrays in the array that contain both x and y
- areAdjacent(x, y) // takes two leaves and returns true if they are adjacent

Added 3 constraints to ctConstraints.js:
- ccPhi(sTree, pTree, adjacent=false) // For every pair (x, y) with some command relation, assign a violation for every ϕ that 
                                                             dominates x but not y, and for every ϕ that dominates y but not x.
- antiCCPhi(sTree, pTree, adjacent=false) // For every pair (x, y) without any command relation, assign a violation if there is no ϕ 
                                                                    containing x and excluding y, and a violation if there is no ϕ containing y and 
                                                                    excluding x.
- mutualSplit(sTree, pTree, adjacent=false) // For every non-mutually-commanding pair (x, y), assign a violation if there is no ϕ 
                                                                       containing x and excluding y, and a violation if there is no ϕ containing y and 
                                                                       excluding x.
- if adjacent == true for these constraints, then violations only count for adjacent words with the specified command relations or 
  lack thereof. This is more reflective of a couple of the Kalivoda (2018) constraints but Kalivoda has expressed uncertainty about that particular specification.

Added files ctConstraints_test.html and ctConstraint_trees.js for testing of additions to ctConstraints.js.


jbellik:
Ed, the interface bit looks good, but let's move the fieldset up to be right under Align/Wrap. Also, I don't think that section necessarily needs a "show more" -- isn't Group just as central to the theory as the other two? I had the impression it was similar to Wrap.

I'm also wondering about the fact that categories are included in the constraint names. Are these supposed to be generalizable to other categories? I did see that the iota can be selected instead of the phi. This makes me think we need to name them something more general, like c command to p-cat (prosodic category). The info text should similarly be adjusted. Also this is probably a question for @nkalivoda , but would it ever make sense to do this for compound words, and select the prosodic word as the category?

Alternately if only the phi option really makes sense, then it would be better to leave off iota and the category selection altogether.

I think we should also discuss how to organize the options. 
- "Require minimal" in particular is a bit opaque right now. I think this means the grouping phi needs to be minimal? Is this a constraint that might be used in conjunction with the more general Group constraint? If so, we should put "Group minimal" as separate constraint in the "show more" section, rather than just as an option under Group.
- Under CC and Group, since there's only one checkbox under "Enforce Command only for syntactic nodes that are...", rather than having a heading + suboptions, let's just have a single line: "Enforce only for adjacent syntactic nodes"

constraints/command.js
```diff
@@ -0,0 +1,486 @@
...
+	return check;
+};
+
+//Assigns a violation for each c-pair whose elements do not reside together in at least one phi.
+function group(sTree,pTree,cat,options)
+{
+	options = options || {};
+	var vcount = 0;
+	var sLeaves = getLeaves(sTree);
+	var phis = getPhis(pTree,cat);
```
jbellik, line 239: This variable phis never gets used.

constraints/command.js
```diff
@@ -0,0 +1,486 @@
...
+	{
+		if(x == myArray[i])
+		{
+			answer = true;
+		}
+	}
+	return answer;
+};
+
+function getNodes(myTree)
```
jbellik, line 19: Is this used anywhere other than in getPhis? If not, I'm not sure it's useful to have around.

jbellik:
In an email, @3ddiehead  asks about how to avoid an error about invalid categories if no categories are included on the interface (if I've understood the question correctly). I think there's a simple way to hardcode "phi" in as the category, which is to set the value to be, for example, group-phi (format: function-cat). Give this a try and let me know if it doesn't work.

@nkalivoda , please weigh in on whether it's desirable to have iota and omega available as categories for the Command Theory constraints, or if we should hard-code phi in on the interface instead.

3ddiehead:
Whenever I try and remove the category options I get an error saying one is needed under current settings.
How can I remove the category options while retaining functionality?

constraints/command.js
3ddiehead: It is used in the "sharePhi" function as well

constraints/command.js
3ddiehead: removed

jbellik (commented):
The interface seems improved, but I'd like to see the checkboxes for options like "Enforce only for linearly adjacent words" on the same line as the text. See: options under "Enforce exhaustivity" in "GEN: Output parameters."

Thanks for removing the redundant function "getPhis". What's the story on the removed "split" functions? Were those also not used anywhere?

3ddiehead:
The split functions are intended to reflect the antiCCPhi constraint but only mark violations for adjacent terminals and for mutual-noncommand situations, not "not mutual command" situations as well.  antiCCPhi allows optionality for both adjacency and the "mutual non" vs. "not mutual" distinction.

jbellik:
I'm trying to ask if you removed those split functions on purpose or accidentally.

Changes to the interface look good, although the linebreak in "mutually non-commanding" is unfortunate.

nkalivoda:
> please weigh in on whether it's desirable to have iota and omega available as categories for the Command Theory constraints, or if we should hard-code phi in on the interface instead.

Apologies for not replying sooner. I don't discuss ι and ω categories in my dissertation, and haven't worked out yet how exactly those constraints would be formulated. I think we should hard-code φ on the interface (for now at least).

jbellik:
Ok, sounds good. I think this branch is ready to go, then. Can you just check that the violation profiles are correct? I will email you some results files.

3ddiehead:
Kalivoda inspired changes made, ready for publication

Issue #442, opened 2020-08-23
MatchSP, MatchPS - combining subcategories with lexical, overtly headed options


nvh4:
Issue with assessing (non-)maximal and (non-)minimal status when either the "lexical" or "overtly headed" option is selected. 
Right now subcategory status is based on the entire tree, but should instead be determined based on the set of lexical and/or overtly headed phrases.

For instance, given the syntax: [<sub>FuncP</sub> X [<sub>LexP1</sub> Y [<sub>LexP2</sub> Z ]]]

When "lexical" is selected, LexP1 should be considered maximal because it is the highest lexical phrase (i.e., it is the highest XP within the set of lexical XPs that are visible to MatchXP-Lex). However, SPOT will consider it to be non-maximal because it is dominated by FuncP (i.e., it is not maximal if one considers the entire tree). Thus, Match(XP-Max-Lex,phi) will incorrectly ignore LexP1. A similar issue arises when "overtly headed" is selected.

Needed: SPOT should first filter out non-lexical (or non-overtly-headed) phrases, and then determine which phrases are max, min, etc.

jbellik:
Please pair program on this as an educational exercise.

jbellik:
We need to add options requireLexical and requireOvertHead to markMinMax(). These can be passed in from MatchSP / MatchPS.

(We won't build support to optionally parse LexP1 as maximal from the example in the original post.)

jbellik:
Proposed implementation:
- create function removeSpecifiedNodes() (modeled on trimRedundantNodes() and maybe also trimSilentTerminals() from treeTrimming.js). It should take the attribute of nodes to be trimmed as an option, and remove whatever nodes have that attribute == true from a deep copy of that tree (use copyTree())
- markMinMax can call removeSpecifiedNodes() with whatever appropriate attributes as the option.
- matchSP will continue to call markMinMax as before but with options.

[ func lex [complementXP] ] --> [lex [complementXP]]
[ func [complementXP] ] --> [complementXP]     <- trimRedundantNodes could handle the transformation from [[complementXP]] to [complementXP]

[.f [subj] [verb] ] --> [.dummy [subj] [verb]]: create some dummy category to replace a functional root to maintain tree status. {id: 'root', cat: 'dummy', children: [{subj}, {verb}]}

3ddiehead:
[...]

PR #453, opened 2020-09-09
Match SP options bug

3ddiehead:
-Bug fixed by modifying markMinMax function to indicate when a node was ignored for being functional&&requireLexical or silent&&requireOvertHead.
-Child can know that its parent was ignored and can become maximal.
-minor logic fix
Testfile is matchMaxMinLex.html


jbellik:
Closes #442

PR #468, opened 2020-09-24
Irish AMP system with SS(phi max)

jbellik:
Nick, let's discuss this set up when we Skype.


jbellik (commented):
- built-in system menu: "Irish x: constraints"
- comment out the Irish AMP systems

jbellik:
Testing revealed that StrongStart(phi max) with evaluation of immediate daughters only as implemented here did not replicate the effects of StrongStartInitial (i). So any built in systems published need to refer to SSInit. However, SS init needed two changes:
- penalize nodes of category cat *or lower* that are initial in a node of category cat+2 *or higher*
- look at all following sisters, not only the immediately adjacent sister, to assess violations

I tried to make both of these changes but have created some kind of bug that is causing SSInit to assign too many violations. @maxTarlov , can you take a look? I don't think I've quite understood what your code is doing.

maxTarlov:
Long story short: the while-loop in the sub-function violation() was not running for nodes with only one child, so the function never falsified the assertion that a unary node incurs a violation of strongStartInit. For this reason, every unary node received a violation when it should not have. Fixed this off-by-one and added check for unary nodes before loop starts.

@jbellik is there a reason the loop ran while i<parent.children.length-1 instead of i<parent.children.length?

jbellik:
A very belated reply -- thanks, Max. That was just a mistake.

Something is still off since {({b (c d)})} is getting 4 violations of strongStartDeep(i) -- a note to myself to look into this further.

jbellik:
Giving up on this constraint / analysis for now since we concluded for theoretical reasons that SS(phi max) was a better implementation.

PR #476, opened 2020-10-18
MatchSP-LexOvert bugfix

3ddiehead:
-Created new function createDummies(tree, attribute) to change the category of nodes with specified attribute to "dummy".
-Modified markMinMax to ignore nodes of category "dummy".
-Generalized trimRedundantNodes to work on functional nodes as well as silent ones. (Not used in bugfix but resulting from the attempted trimming strategy suggested in issue #442)


constraints/recursiveCatEvals.js
```diff
@@ -5,18 +5,29 @@ Assumes all nodes have valid and relevant categories
 (i.e., this is designed for prosodic trees and won't give the desired results
 if run on a syntactic tree that contains, e.g., bar levels).
 */
-function isMinimal(node){
-	var cat = node.cat;
+function isMinimal(node, lastCat){
+	//use the lastCat argument as the cat under question if it exists.
+	if(lastCat){
```
jbellik, line 10: Please expand on this comment to explain what lastCat and when it would be defined. Or you can put that information in the comment above the function declaration.

constraints/recursiveCatEvals.js
```diff
@@ -5,18 +5,29 @@ Assumes all nodes have valid and relevant categories
...
 	if(!node.children || !node.children.length)
 		return isMin;
 	//Otherwise, we have to look at its children to determine whether it's minimal.
 	var i = 0;
 	var chil = node.children;
 	while(isMin && i<chil.length){
-		if(chil[i].cat===cat)
+		//if a child is a dummy, we will have to see skip over that dummy to see if any of its children have the same category.
+		if(chil[i].cat == "dummy"){
+			isMin = isMinimal(chil[i], cat)
```
jbellik, line 27: This must be where lastCat is relevant.

constraints/recursiveCatEvals.js
```diff
@@ -56,26 +67,53 @@ var sCat = ["cp", "xp", "x0"];
  * 7/29/19 refactor of an earlier version
```
jbellik, line 67: Please expand on this comment to explain the definitions of minimal and maximal we're working with when requireLexical and requireOvertHeads come into play.
Also provide a list of options in the comment before the function.

constraints/recursiveCatEvals.js
```diff
@@ -56,26 +67,53 @@ var sCat = ["cp", "xp", "x0"];
...
+	if(mytree.cat == "dummy"){
+		mytree.isMax = false;
+		mytree.isMin = false;
+		if(mytree.parentCat != "dummy"){
+			mytree.lastCat = mytree.parentCat;
+		}
+	}
+
+	//recall stored parentCat after dummies are skipped
+	if(mytree.parentCat == "dummy" && mytree.cat != "dummy"){
```
jbellik, line 102: I think this should be '===' instead of '==' and '!==' instead of '!=' (strict equality). Not sure it actually has an effect though.

jbellik, line 102: Line 93 deals with the case where mytree.cat == "dummy", and line 102 here deals with the case where mytree.cat !== "dummy" AND mytree.parentCat !=="dummy". But do you ever deal with the case where mytree.cat !== "dummy" but mytree.parentCat == "dummy"? is that relevant?

I think this line should probably be an else if, in fact, to more clearly link it to the preceding conditional (line 93):
else if(mytree.parentCat !== "dummy"){...}

jbellik: Or are they not supposed to be linked? Now I'm doubting that i've parsed this code correctly.

constraints/recursiveCatEvals.js
```diff
@@ -56,26 +67,53 @@ var sCat = ["cp", "xp", "x0"];
...
+		mytree.parentCat = mytree.lastCat;
+	}
 
-	//mark minimality (relies on isMinimal above)
-	mytree.isMin = isMinimal(mytree);
+	//mark maximality and minimality for node
+	if(mytree.cat != 'dummy'){
```
jbellik, line 107: Possibly you should nest both this and the conditional in line 102 under a big "else" statement (the "else" of "mytree.cat === "dummy" in line 93) so you don't have to separately check for dummy-hood repeatedly.

jbellik (changes requested):
I'm not quite done reviewing this, and will need to look closely at all the tableaux to think through whether the violations are correct. But I'd like to ask you to add comments to all of this code to clarify what is going on, since there are theoretical decisions being made / enforced in this code (as evidenced by the discussion on the issue), and it would be good to have documentation within the code itself reminding us what those decisions are. Comments will also help me review it better.

jbellik:
Closes #442

constraints/recursiveCatEvals.js
```diff
@@ -53,29 +86,56 @@ var sCat = ["cp", "xp", "x0"];
...
 	}
 
-	//mark maximal nodes
-	mytree.isMax = (mytree.cat !== mytree.parentCat);
+	//Store the info of the most recent cat in order to skip over dummy nodes
+	//except when the dummy node's parent is ALSO a dummy node, then lastcat should be passed
+	//down dummy generation after dummy generation until a normal node is reached to inherit
+	//it as the parentCat.
+	if(mytree.cat === "dummy"){
+		mytree.isMax = false;
```
jbellik, line 118: I'm not sure there's any reason to label max/min on dummies, but I don't think it does any harm either.

constraints/recursiveCatEvals.js
```diff
@@ -53,29 +86,56 @@ var sCat = ["cp", "xp", "x0"];
...
+		mytree.isMin = false;
+		if(mytree.parentCat !== "dummy"){
+			mytree.lastCat = mytree.parentCat;
+		}
+	} else {
```
jbellik, line 123: This flow of logic is much clearer than before. Good work.

constraints/recursiveCatEvals.js
```diff
@@ -53,29 +86,56 @@ var sCat = ["cp", "xp", "x0"];
...
+		}
+	}
 
 	if(mytree.children && mytree.children.length){
 		for(var i = 0; i < mytree.children.length; i++){
-			var child = mytree.children[i];
-			child.parentCat = mytree.cat; // set the property parentCat
-			mytree.children[i] = markMinMax(mytree.children[i]);
+			mytree.children[i].parentCat = mytree.cat; // set the property parentCat
+			mytree.children[i].lastCat = mytree.lastCat; //pass on lastCat
```
jbellik, line 137: Given that you're recording lastCat on all the nodes, I'm not clear on why it's also needed as an argument to isMinimal. Can't isMinimal refer to node.lastCat? But perhaps I'm missing something.

treeTrimming.js
```diff
@@ -46,6 +46,31 @@ function trimSilentTerminals(inputTree){
 	return trimSilentInner(treeCopy);
 }
 
+/* function removes non-lexical heads from a tree. Basically identical to trimSilentTerminals
```
jbellik, line 49: Am I right in thinking that markMinMax() doesn't ultimately use this? Do we actually need this function after all? Sorry, I'm sure we talked about this when you started on this issue, but I don't remember the details.

jbellik: In fact, isn't this function redundant with the more general trimAttributedNodes()?

treeTrimming.js
```diff
@@ -99,3 +128,63 @@ function trimRedundantNodes(inputTree){
 	}
 	return trimInner(tree);
 }
+
+function trimAttributedNodes(inputTree, attribute){
```
jbellik, line 132: I think trimNodesWithAttribute() would be a better function name.

treeTrimming.js
```diff
@@ -99,3 +128,63 @@ function trimRedundantNodes(inputTree){
...
+					node.children[i] = createDummyInner(child);
+				}
+			}
+		}
+		return node;
+	}
+	return createDummyInner(tree, attribute);
+}
+
+function removeSpecifiedNodes(inputTree, attribute){
```
jbellik, line 183: Does this actually ever get used, or is it left over from a solution path that we abandoned?

jbellik (changes requested):
The markMinMax() code looks great. Thanks for adding those comments and clarifying the conditional structure.

I am not sure all the new code in treeTrimming.js is actually still relevant. See comments on specific lines. If there is code left over from abandoned solutions, please delete it, comment it out, or at least indicate in comments that it is not being used.

It looks to me like everything is working nicely. The last thing i'd like to see happen before merging is: add Mocha tests to test/matchMarkMaxMinLex.html! I found myself mentally checking that the values are correct and it would be best to go ahead and add those tests now before merging, just in case anything troublesome turns up.

treeTrimming.js
jbellik: This code was not written by Ed, possibly was written by Max. Not used by markMinMax(). Unclear whether it works or is used anywhere.

treeTrimming.js
jbellik: This is not used anywhere but Ed says it should work.

treeTrimming.js
jbellik: This is used in markMinMax().

jbellik: Upshot of conversation with Ed: This is not getting called anywhere, but it could be useful someday. It is indeed leftover from an abandoned strategy.

jbellik (approved):
This will need to have Mocha tests added but I'm merging it as-is for now.

Issue #470, opened 2020-09-24
Pitch accent constraints requiring tone-marking in built-in system


nkalivoda:
In the Ito & Mester built-in system, accentAsHead and noLapseL stop assigning violations when I uncheck "Annotated with tones". They should still work without the annotations.

jbellik:
Diagnosis: accentAsHead and noLapse are counting on the ptrees having accent attributes in them. Originally if accent attributes were missing, they would be inferred from the node ids (a --> accent: true, u --> accent: false). but this isn't working now. Possibly due to a change in GEN?

Make the accent constraints more robust by having them call another function to add accents as necessary

jbellik:
Define a function that takes a tree, and if any of its nodes has an id of the form:
'a'
'A'
'a_#'
'A_#'
where # = any integer,
then give it an attribute accent set to true.

Otherwise, accent can be set to false, or left unassigned.

PR #478, opened 2020-10-20
Accent bugs

jbellik:
Fix bugs with accentFromId() which were impacting accentAsHead and noLapseL. Also minor fixes to wrap() and to addJapaneseTones() in annotate_tones.js.


jbellik:
Closes #470

PR #487, opened 2020-10-26
issue #463 trees with lots of terminals

M130:


tijgee:
changed treeUI-tree display to inline-block

jbellik (approved):
Great, thanks for figuring this out!

PR #495, opened 2020-11-02
Match Automated Testing

maxTarlov:
@M130 and I finished turning @nvh4 's test methods into automated tests using mocha. There is still some naming of tests to occur, but personally I would like feedback before editing all of the names.

Proposed naming scheme: describe blocks are labeled according to their corresponding tableau in @nvh4 's file, followed by a colon, followed by the syntax parameters (if any) that are being held constant in that test block. It blocks are named according to the varied prosody parameters.

Example:
## Tableau 10: nonMinSyntax
 - [X] maxProsody
 - [X] nonMaxProsody
 - [X] minProsody
    ...

### DIFF
--- a/test/_test_template.html
+++ b/test/_test_template.html
@@ -1,0 +1,85 @@
+<html><head> 
+    <title>SPOT Automated Testing</title>
+
+	<link rel="stylesheet" type="text/css" href="../spot.css">
+
+    <!--If running tests that do not have to do with the interface, run jsbuild.sh or jsbuild.bat
+        in root directory of the project. If running interface testing, run testbuild.sh
+        or testBuild.bat-->
+    <script src="../build/spot.js"></script>
+
+    <!--Mocha setup stuff (1/2)-->
+    <meta charset="UTF-8">
+    <script src="../lib/test/mocha.min.js"></script>
+    <script src="../lib/test/chai.min.js"></script>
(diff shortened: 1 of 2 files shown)


jbellik:
This looks great! I agree the names need some work for transparency. A first step will be to include the bracket representation of both the stree and ptrees in the text for each it(), I think. Or if the trees are the same within each block, the bracket representations could go in the describe() statements.

jbellik:
Other naming improvements can be made too but let's discuss them during the meeting tomorrow

maxTarlov:
> A first step will be to include the bracket representation of both the stree and ptrees in the text for each it(), I think. Or if the trees are the same within each block, the bracket representations could go in the describe() statements.

The same stree is used for every assertion, but each "it" block uses at least two ptrees, and most use about 6. Maybe instead of putting the bracket representation of all 3-6 trees in the it block, we can show the trees in the message of each assertion. In this case, you would only see a tree if an assertion fails. 

(If you look at @nvh4 's tests, each "describe" block is a tableau and each "it" block is a constraint column in that tableau.)

jbellik:
That sounds like a good solution.

maxTarlov:
I have implemented a version of the above proposal for Tableau 1. To see it in action, pick an arbitrary assertion in the first describe block and change the expected value.

jbellik:
Add README type info at the top, including references to Nick Van Handel's files & possibly blocks of his text from the pdf file.

Issue #479, opened 2020-10-21
Automated testing for Binarity constraints


jbellik:
Create an html document that uses Mocha tests to check the values produced by all the non-deprecated binarity functions in the file main/constraints/binarity.js. You can assume that the values currently being produced on the interface are correct.

Some sample expected values:

For ptree = ((a b)(c d)), 
- binMaxLeaves({}, ptree, 'phi') = 1
- binMaxBranches({}, ptree, 'phi') = 0
- binMinLeaves({}, ptree, 'phi') = 0
- binMinBranches({}, ptree, 'phi') = 0

For ptree = (((a) b) c) d), 
- binMaxLeaves({}, ptree, 'phi') = 2
- binMaxBranches({}, ptree, 'phi') = 0
- binMinLeaves({}, ptree, 'phi') = 1
- binMinBranches({}, ptree, 'phi') = 1

For ptree = {a b c d} (i.e., the root is of cat: 'i'), all functions with category argument = 'phi' should return 0.
- binMaxLeaves({}, ptree, 'phi') = 0
- binMaxBranches({}, ptree, 'phi') = 0
- binMinLeaves({}, ptree, 'phi') = 0
- binMinBranches({}, ptree, 'phi') = 0

For ptree = ((a b c)(d e f)),
- binMaxLeaves({}, ptree, 'phi') = 3
- binMaxBranches({}, ptree, 'phi') = 2
- binMinLeaves({}, ptree, 'phi') = 0
- binMinBranches({}, ptree, 'phi') = 0

For ptree = {(a)(b)(c)},
- binMaxLeaves({}, ptree, 'phi') = 0
- binMaxBranches({}, ptree, 'phi') = 0
- binMinLeaves({}, ptree, 'phi') = 3
- binMinBranches({}, ptree, 'phi') = 3

These samples were not chosen all that systematically; you should try to think about which ones are the most useful and include those, and also consider if there are other edge cases that haven't been covered here.

PR #501, opened 2020-11-09
Binarity test

tijgee:


jbellik (changes requested):
This is a good start. There are a few changes I'd like to see:

1. Currently, each it statement has 4 asserts in it, each of which refers to a distinct constraint function. It would be better for each it to only refer to one constraint function. That way if the test fails, it's more obvious what has gone wrong. We don't want it to be ambiguous which function is causing the problems. My suggestion is for each describe statement to refer to a particular tree. You can put parenthesizeTree(ptree1) (or whatever tree it is) in the text part of the describe statement.

2. The text in the describe statements and the it statements needs to be more descriptive. It should refer to the function being tested and include the bracket representation of the tree being tested as well. Right now I can't tell what the tests are doing unless I look at the code, but i should be able to look at the html in the browser and see what is and isn't being tested.

jbellik (approved):

jbellik:
Closes #479

PR #507, opened 2020-11-16
Input validation min input and max input

M130:
Input validation min input and max input


jbellik (commented):
This is looking good. I made some minor changes to the wording of the errors.

I can see that this validation ended up being fairly complicated. Can you add some comments to clarify the flow of logic for future readability? In particular, it would be helpful to add a comment stating the purpose of the variables mintest and maxtest; I think these are essentially flags for whether other tests need to happen, but since they are integers I initially thought they were counting something.

It would also be good to add a check that "List of terminals" is non-empty.

I wish I had thought of this sooner, but it occurred to me that it might be simpler to do some of this input validation by changing the input fields to be drop-down menus with values from 1 to 10. Then we wouldn't need to check that they are numbers or check that they are greater than 0 and less than 10. It might even work to disable the max values that are less than the selected min value.

We can discuss this idea tomorrow if you like.

M130:
I can add the comments and check for List of terminals being non-empty. As for the drop down menus, I think that's a good idea. We can discuss it more tomorrow.

M130:
Added the comments and test case for list of terminals.

maxTarlov:
Unit Tests for save/load have been merged into master, no need to merge from save_load_test
To run unit tests on this:
* run testBuild.sh
* open interface1.html
* in JS console, use `runStringTest()`
* mocha reporter should appear at the top of the page
N.B. the unit tests right now use numbers out of the range that you have defined. If that makes the test fail, you can change them in the unit test code, i am working on a better way to check that things have saved

jbellik (changes requested):
I made some minor adjustments to the messages and the variable names. The one thing that doesn't seem to be working quite right is that the check for having a string of terminals fails any time one of the "List of terminals" fields is empty. But it should only fail if *all* of them are empty. There isn't a way to delete the additional lists of terminals once you click "Add list of terminals" so we don't want to require the user to fill in every single list of terminals.

I hope that makes sense; let me know if you have any questions.

M130:
If there's no terminal input then do you need the max and min input for that terminal?

M130:
I changed the code so it would work if there is at least on list of terminals input. There is no input needed for max and min if the list of terminals is empty.

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
...
+										minmax = 1;
+									}
+								}
+							}
+							if (minmax === 0){
+								/* display warning if the input value is more than 5 */
+								if (minProblem === 1 || maxProblem === 1){
+									minProblem = 0;
+									maxProblem = 0;
+									displayWarning("Min or Max input is greater than or equal to 5 which may cause your browser to freeze due to too many terminal strings being generated.");
```
jbellik, line 1024: This warning is displaying too late. The calculation of terminal strings runs before the warning appears, so the browser seems to just become unresponsive with no warning. Can you actually convert it to a confirm() dialog, so the user has a chance to change their minds before they freeze / lag their browser? You may also need to move it to a different spot in the logic to get it to appear in time.

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
...
+					}else{
+						if (isNaN(spotForm.genStringsMin.value)){
+							minProblem = 1;
+						}
+						if (isNaN(spotForm.genStringsMax.value)){
+							maxProblem = 1;
+						}
+					}
+				}
+				if (minProblem === 0 && maxProblem === 0){
```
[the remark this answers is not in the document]
jbellik: I guess the reason for the current organization is to be able to give more informative errors.

How about flipping the logic of the conditionals, so that instead of nesting them all, it goes:
if(input is non empty but min is empty or max is empty){minOrMaxProblem = 1; display appropriate error}
elseif(isNaN(min) or isNaN(max)){minOrMaxProblem = 1; display a different appropriate error}
elseif(other checks...)
else //if you get here, there are no problems with min or max

But perhaps I'm overlooking something and the current organization is unavoidable.

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
...
+								}
+							}
+							if (minmax === 0){
+								/* display warning if the input value is more than 5 */
+								if (minProblem === 1 || maxProblem === 1){
+									minProblem = 0;
+									maxProblem = 0;
+									displayWarning("Min or Max input is greater than or equal to 5 which may cause your browser to freeze due to too many terminal strings being generated.");
+								}
+								var length = spotForm.genStringsInput.length;
```
jbellik, line 1026: Isn't there already code doing this in lines 890-893? I think 1026-1029 is redundant.

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
...
+								if(length === undefined) {
+									length = 1;
+								}
+								var inputList = spotForm.genStringsInput.value;
```
jbellik, line 1030: Possibly 1030-1032 is also redundant?

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
...
-				inputList = inputList.trim().split(' ');
-				var currGenStringsList = generateTerminalStrings(inputList, min, max)
-				displayStringsTable(currGenStringsList);
 
-				if(genStringsList) {
-					genStringsList = genStringsList.concat(currGenStringsList);
-				}
-				else {
-					genStringsList = currGenStringsList;
+								for(var i=0; i<length; i++){
```
jbellik, line 1035: Add a comment indicating that this is where the actual calculation of the terminal strings happens.

jbellik: I'm hoping that with the flipped logic I suggested above, this chunk would be in the "else" block (where no errors had been encountered), instead of being embedded in half a dozen nested if statements.

jbellik (changes requested):
Thanks for that fix. It's working nicely. 

I think all the functionality is good now, except for one issue which is the timing of the warning about max > 5. 

I did a close read of the code organization, and I think it can be simplified in several ways that will help with readability and maintainability. Specific comments are attached to individual lines of code. It may be that I've suggested some untenable changes; if so, please let me know!

I'm impressed that you managed to get all the logic working with such a complicated structure -- good job!

interface1.js
```diff
@@ -884,35 +884,200 @@ window.addEventListener('load', function(){
 			genStringsList = fixedStringList;
 		}
 
+		/* minProblem, maxProblem, and inputProblem are flags for specific tests. The default value is 0. If the value equals to 1, then the test failed */
+		var minProblem = 0; var maxProblem = 0;
+		var inputProblem = 0;
 		var length = spotForm.genStringsInput.length;
```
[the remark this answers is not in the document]
M130: the length variable to numTerminalStrings?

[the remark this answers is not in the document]
jbellik: Yes, sorry, I was referring to the length variable.

interface1.js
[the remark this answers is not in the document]
M130: Yes, the current organization was to be able to give more informative errors. I will try the suggestion, I think that will work better. Thanks!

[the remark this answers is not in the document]
jbellik: Sounds good!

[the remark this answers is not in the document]
M130: After rethinking, I am not sure if that will work because there needs to be a for loop before each statement that checks all the string. On second thought, I think I structure my code like that previously but that only work for the first input and it didn't work for input 2 and onward so I switched it to the current structure. I will try to rethink about it again though.

interface1.js
M130: I will switched it to confirm statement.

interface1.js
M130: inputList, min, max?
That part is necessary.

interface1.js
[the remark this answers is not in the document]
M130: Ok, got it. Thanks!
That was actually one of the original code so I think if I switch it, I probably need to change other code that wasn't written by me. Do you want me to change them?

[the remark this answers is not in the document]
jbellik: yep, that's fine! I think I just didn't check the readability of Iva's code well enough when she first did this, or it seemed readable enough at the time when there was less code dealing with this feature.

interface1.js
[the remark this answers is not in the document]
jbellik: Hmm. I was thinking that all of the checks could be in one for-loop. But maybe i'm missing something. You can let me know tomorrow.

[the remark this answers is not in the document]
M130: I figured out how to do it in one for-loop. I think it's less complicated and less code.

M130:
Thank you. I think it's less code and less complicated now. I hope it's more readable now.

M130:
There is also a save_load branch I did for the drop-down option.

interface1.js
```diff
@@ -884,35 +884,154 @@ window.addEventListener('load', function(){
...
-
-		for(var i=0; i<length; i++){
-			if(length > 1) {
-				inputList = spotForm.genStringsInput[i].value;
-				min = spotForm.genStringsMin[i].value;
-				max = spotForm.genStringsMax[i].value;
+		/*if inputPresent is 0 then all the List of terminals are empty else if 1 then there
+		is at least one terminal input*/
+		if(inputPresent == 1){
+			breakloop:
```
jbellik, line 909: rename to terminalStringsValidationLoop

interface1.js
```diff
@@ -884,35 +884,154 @@ window.addEventListener('load', function(){
...
+			for(var i=0; i<numTerminalStrings; i++){
+				if (numTerminalStrings > 1){
```
jbellik, line 911: add variables:
input, min, max that are set to either spotForm.genStringsInput (if numTerminalStrings ==1 ) or spotForm.genStringsInput[i] if numTerminalStrings > 1, and so on.
Then eliminate the if/else structure here with its code duplication.

M130:
As I was doing #510, I realized that there was a bug for the function in general. If there is no input for the list of terminal even when the generate combinations and permutations tab is closed, there was still an error staying at least one list of terminal input was needed. It doesn't affect anything but the error display should be removed. I was looking through the interface1.js and I think that the function genTerminalStrings() should break up into two functions or duplicate some of the code from that function in autoGenInputTree() and not call the function. I think you should take a look at the that function genTerminalStrings() and function autoGenInputTree().

jbellik:
Hmm, good catch. I will take a look. That code was originally written by someone who's not on the team anymore, so I'm not super familiar with it.

interface1.js
```diff
@@ -884,35 +889,127 @@ window.addEventListener('load', function(){
 			genStringsList = fixedStringList;
 		}
 
-		var length = spotForm.genStringsInput.length;
-		if(length === undefined) {
-			length = 1;
+		var inputIsFive = false; //if the min or max input is 5 flag
+		var minOrMaxProblem = false; //if there is a min or max input problem flag
+		var inputPresent = 0; //if list of terminal input is present flag
+		var problemError = ""; //string indicating what the min or max problem is
```
jbellik, line 895: This variable doesn't seem to get used later.

jbellik:
I see what you mean. I got started on this but wasn't able to finish this afternoon, so let's discuss how to proceed tomorrow.

jbellik:
I fixed the error. 

But there's a question we need to address, which is: What are the conditions for us to run generateTerminalStrings() and do all its input validation? In particular, should we still run it as part of what "Generate trees" does even if the fieldset for "Generate combinations and permutations" is closed? That's the current behavior. I had put in a line of code that checks whether the fieldset is open before running the terminal gen, but I commented it out because I was worried that might be an opaque behavior.

A related question: Under what circumstances do we want to display an error because no inputs were provided to terminal gen? I changed the code so that this error will only display if they click the "Generate terminal strings" button without providing any inputs. There is still a warning in the console if they click "Generate trees" and there is nothing in the combinations/permutations input fields, but we could get rid of this or decide it should only appear in some situations, etc.

interface1.js
jbellik: renamed to correspond with uses later on.

jbellik (approved):
Changes made by me.

PR #508, opened 2020-11-17
Logging2mocha

tijgee:


test/binMaxHeadTest.html
```diff
@@ -178,6 +205,30 @@
     }
 
 var s = null;
+
+describe('Automated Test', function() {
+            it('binMaxHead(s, x, w)', function() {
```
jbellik, line 210: Add parenthesizations of the prosodic trees (x, y, z, a) to the display text of the it() statements, and also the expected violations.

test/binMaxHeadTest.html
```diff
@@ -178,6 +205,30 @@
     }
 
 var s = null;
+
+describe('Automated Test', function() {
```
jbellik, line 209: Add the constraint name and possibly also the definition (you can get this from the comments in the file that defines the function, in this case most likely constraints/binarity.js) to the displayed text of the describe() block.

test/binMaxPhiMinTest.html
```diff
@@ -123,6 +147,37 @@ <h2>Binarity subcat</h2>
 
 var s = undefined;
 var c = 'phi';
+
+
+describe('Tree1: ( chihoo [[ken satsu] choo] ) = ( w [w F] ) ', function() {
+            it('binMax_minLeaves(s, tree1, c)', function() {
```
jbellik, line 153: Add expected violations to the it() text

test/containsAdjunctTest.html
```diff
@@ -103,11 +129,31 @@
...
         ]
     }
-
 console.log('[[a] b [c]]')
 console.log('expected: false')
 console.log('actual: ' + containsAdjunct(z))
 
+//automated test
+describe('[a [b [c]]]', function() {
+            it('containsAdjunct(x)', function() {
```
jbellik, line 138: Restate these to say "does not contain an adjunct" if the expected value is false, or "contains an adjunct" if the expected value is true.

test/gen_terminal_strings_test.html
```diff
@@ -1,6 +1,6 @@
 <html>
 <body>
-<script src="../../build/spot.js"></script>
+<script src="../build/spot.js"></script>
 
 <script>
```
jbellik, line 5: These could be converted to Mocha tests too.
For example:
describe('terminal strings drawn from [a, b, c]', function(){
   it('with minimum length 2, maximum length 2', function(){
       assert(....)
   }
}
and so on. You can assume that the currently generated sets of terminal strings are correct in setting up the expected values.

test/headAlignBugTest.html
```diff
@@ -343,6 +370,58 @@
 var strict = "strict";
 // strict = undefined;
 
+//automated test
+describe('goodTree1: ' + parenthesizeTree(goodTree1), function() {
```
jbellik, line 374: - include the expected values in the displayed text for the it() statements
- specify the value of side -- it should be either 'right' or 'left', i think. I'm not sure where it's being defined at present.

test/mirror_test.html
```diff
@@ -1,6 +1,32 @@
...
+<body style="padding-left: 5%; padding-right: 5%; padding-top: 20px">
+    <h2>SPOT Automated Testing</h2>
+    <script>
+        mocha.setup("bdd"); //brings "describe", "it", etc. into global namespace
+            //mocha.checkLeaks();
+            var assert = chai.assert; //no one wants to type out "chai.assert" every time
+    </script>
+    
+    <div id="mocha">
+        Head Align Automated Test
```
jbellik, line 28: This is a test of whether trees that are mirror images of each other are getting properly filtered out, not of head alignment

test/mirror_test.html
```diff
@@ -236,6 +262,55 @@
 console.log("expected: ", false)
 console.log("\n")
 
+//automated test
+describe('checkMirror: ', function() {
```
jbellik, line 266: Please copy in comments on what checkMirror and mirrorImages are supposed to do, because the names aren't really transparent. These are defined in main/syntaxFilters.js.

test/noShiftGradientTest.html
```diff
@@ -0,0 +1,134 @@
...
+    porderb = ["b", "a", "c", "d", "x"];
+    console.log("b a c d x: ", noShiftGradient(sorder3, porderb), " expected: 1")
+    porderc = ["c", "a", "b", "d", "x"];
+    console.log("c a b d x: ", noShiftGradient(sorder3, porderc), " expected: 2")
+    porderd = ["d", "a", "b", "c", "x"];
+    console.log("d a b c X: ", noShiftGradient(sorder3, porderd), " expected: 3")
+
+    //automated test
+    describe('Moving X,  "a b c d X"', function() {
+            it('"a b c d X:", noShiftGradient(sorder, porder)', function() {
```
jbellik, line 72: Would it work to declare all these variables closer to where they are being used? I think that would improve the interpretability of this file.

test/threeXPsTest.html
```diff
@@ -1,6 +1,32 @@
...
+  </head>
+  <h2>SPOT Automated Testing</h2>
+  <script>
+    mocha.setup("bdd"); //brings "describe", "it", etc. into global namespace
+        //mocha.checkLeaks();
+        var assert = chai.assert; //no one wants to type out "chai.assert" every time
+</script>
+
+<div id="mocha">
+    threeXPsTest
```
jbellik, line 27: It would actually be best to combine this with the containsAdjunct test file. These are closely related functions.

jbellik (changes requested):
This is a very good start. I've made a number of individual comments on how readability of the test files can be improved. 
Eventually it would be nice to combine some of these test files, because several of them deal with syntax filters that are all used by sTreeGEN(). But that can happen on a later round of revisions.

test/noShiftGradientTest.html
```diff
@@ -0,0 +1,134 @@
...
+    console.log("X d c b a: ", noShiftGradient(sorder3, pordera), " expected: 10")
+    porderb = ["b", "a", "c", "d", "x"];
+    console.log("b a c d x: ", noShiftGradient(sorder3, porderb), " expected: 1")
+    porderc = ["c", "a", "b", "d", "x"];
+    console.log("c a b d x: ", noShiftGradient(sorder3, porderc), " expected: 2")
+    porderd = ["d", "a", "b", "c", "x"];
+    console.log("d a b c X: ", noShiftGradient(sorder3, porderd), " expected: 3")
+
+    //automated test
+    describe('Moving X,  "a b c d X"', function() {
```
jbellik, line 71: change the message here to "sorder =  a b c d X"

test/noShiftGradientTest.html
jbellik: Declare / define porder inside the it() block so that it's easy to see what its value is in the browser.

test/noShiftGradientTest.html
```diff
@@ -0,0 +1,134 @@
...
+        //mocha.checkLeaks();
+        var assert = chai.assert; //no one wants to type out "chai.assert" every time
+</script>
+
+<div id="mocha">
+    No Shift Gradient Automated Test
+</div>
+  <script>
+    // example with moving x
+    console.log("a b c d X")
```
jbellik, line 33: All the console logging can be deleted since it is replaced by the Mocha tests.

jbellik (changes requested):
This is good progress! See specific comments for more things that I would like clarified.

Ideally, I'd also like to see  mirror_test, headAlignTest, and containsAdjunct / threeXPsTest combined into a single file: syntacticFiltersTest.html

Also, can you convert filenames that use underscores to be camel cased instead? That seems to be the more common naming scheme and it would be nice for these to be uniform.

Thanks!

PR #529, opened 2020-12-07
warning when both manual and automatic input have content added

M130:


interface1.js
```diff
@@ -421,14 +421,22 @@ window.addEventListener('load', function(){
...
-
 		// Get the code that is in the stree textarea
 		var treeCode = spotForm.sTree.value
 		// if code has been generated, then ignore pString in GEN
 		if(treeCode !== "{}") {
 			pString = "";
 		}
-
+		if(document.getElementById('inputOptions').style.display == 'block') {
+			if (spotForm.inputToGen.value != ""){
```
jbellik, line 431: We also need to check if there is content in the code version of the list of trees.

jbellik: I just implemented this since it was a quick fix.

interface1.js
```diff
@@ -421,14 +421,23 @@ window.addEventListener('load', function(){
...
 		}
-
+		var doubleInputWarningMsg = "Inputs were provided on both the Manual tab and the Automatic tab of Gen: Inputs. The candidate set will be created using inputs on the tab that is currently visible. Inputs that are not currently displayed will be ignored.";
+		if(document.getElementById('inputOptions').style.display == 'block') {
+			if (spotForm.inputToGen.value != "" || treeCode != "{}"){
+				displayWarning(doubleInputWarningMsg);
+			}
+			pString = "";
+		}else{
+			if (spotForm.inputToGenAuto.value != ""){
```
jbellik, line 437: This is not the best field to check the value of, because it's just the first terminal string field, rather than the actual list of trees. Try checking the value of getAutoStreeList() instead.

The problem to address: If the user makes some trees in the Automatic tab, then deletes the list of terminals out of inputToGenAuto for some reason, then build a tree or trees in the Manual tab, and clicks Get results, then the warning does not display.

jbellik (changes requested):
Pretty good! I made a couple simple changes; see comments on the code. There is one more thing I'd like you to change though, which has to do with what you're checking to see if there is content in the Automatic tab.

jbellik (approved):

PR #539, opened 2020-12-10
Refactor interface1

### LINKED ISSUE #512: Refactor interface1.js
This file's length and complexity are getting out of control. Let's try to split it up into at least the following files:
- interface_terminalStringGeneration.js: all the logic around generating lists of terminal strings
- interface_treeBuilder.js: code that goes with the manual tree-builder
- interface_autoStrees.js: code concerning automated generation of syntactic trees and the many parameters for that
- interface_makeTableau.js: code that actually makes the tableau, and provides warnings if there are going to be too many candidates
- interface_utilities.js: functions like closeButton(), displayError(), displayWarning(), saveAs(), and possibly others

I haven't evaluated whether some of this might be hard to disentangle; if they are, we can re-evaluate this particular division of the code.

jbellik:
- splits up interface1.js into several smaller files, and creates lots of smaller, named functions in the process
- puts interface-related JS files into a subdirectory
- amends jsbuild.sh to account for this

testBuild.sh and build.bat have not been updated yet, however!

Closes #512


jbellik:
@nkalivoda , @maxTarlov : I believe everything is still working based on my many tests and fixes as I went along, but in the absence of systematic automated testing, it would be helpful to have you both try out some things in case you run into any issues i missed.

I will fix the other build scripts eventually before merging this pull request.

PR #543, opened 2021-01-09
Align wrap testing

3ddiehead:
Align Wrap testing tentatively completed:
- Tests all option combinations for standard Align function (Left and Right)
- Tests Wrap function


jbellik (changes requested):
I think these look pretty good but they do need a little more informational text. At the top of the Align and Match auto test files, where it says "This is where test results should show up", please add a couple sentences explaning where the tableau numbers are coming from and what the methodology was for setting up these tests. I'm guessing the numbers refer to the tableau numbers in Nick Van Handel's doc? If the methodology is also based on NVH's doc, you can just state that that's where everything is explained and include a link or links.
It would also be helpful if there was an easy way to see what the actual trees are... Is there a different document where the tableaux actually exist? If so, perhaps the two documents could be combined?

jbellik:
Upshot of conversation 1/14: 
- there are no tableaux so just remove references to tableaux
- grab Max's custom error function off of the auto Match testing instead of worrying about displaying all of the trees
- put parenthesizeTree(stree1) in description at top of page.

3ddiehead:
Done here.

jbellik (approved):

PR #545, opened 2021-01-16
jsbuild.bat modified

3ddiehead:
-Added line to build files in interface_JS folder into build/interface.js
-Changed line to include interface.js in spot.js

Looks like it works? All the files seem to be properly built into spot.js


jbellik (approved):
I don't have a good way to test that this is working on my Mac, but it looks right, so I will trust your assessment that all the files have compiled correctly into spot.js. Thanks!

PR #549, opened 2021-01-22
Created Automated GEN Testing

3ddiehead:
One question: I mention "Candidate Set Cardinalities in SPOT" as a reference for parameters but I'm not sure the best way to link to that or if it's even the best way to reference the info.  Should I just write a more thorough description of the parameters?


jbellik (commented):
Let's have links to the spreadsheet and to the Google doc of the paper. We can eventually replace that with a link to a pdf in this repository when we finalize the paper. Please also copy and paste in the paper's descriptions of the parameters. Otherwise, this looks good to go! Next step will be creating tests for GEN with clitic movement.

jbellik (approved):

PR #552, opened 2021-02-01
GEN Clitic and Permutation testing

3ddiehead:
Finished the tests, should I add a description of GENwithCliticMovement/Permutation or link to any documentation?

Also, I can calculate the formulae for a couple of the clitic movement ones that aren't just triple the normal.


jbellik:
This looks good. I don't think we have good documentation for the reordering GENs, unfortunately.

PR #553, opened 2021-02-02
Automated test string input validation

M130:
Mocha tests are working without any changes to the interface1 but can't really determine which tests the warnings and error displays are coming from because all the tests are ran in an instant and there is no assert to check if there is a display warning or not. Please check it to see if the current tests will work or if they need to be changed.


test/stringInputValidationTest.js
```diff
@@ -0,0 +1,117 @@
...
+}
+
+function runStringInputValidationTest() {
+    setUp();
+    describe("Generate trees", function() {
+        it("Generate combinations and permutations not added - no warning", function() {
+            document.getElementById("spotForm")["genStringsInput"].value = "";
+            document.getElementById("spotForm")["genStringsInput"].length = undefined;
+            document.getElementById("stringGeneration").classList = [""];
+            genTerminalStrings(); //add that generate tree button was clicked instead
```
jbellik, line 23: Why "instead"?

jbellik (changes requested):
Good progress! I'm glad you were able to figure out whatever the issue was with set up on the interface last week. The set of cases being tried out looks good too. So this is already quite useful, I think!

When I ran this, I got a series of pop-ups: "This page says: [warning message from SPOT]. Confirm to continue." So that enabled me to see that the various tests were being executed, and sort of tell that there error/warning messages were popping up, although I couldn't easily tell if they were the correct ones.

I don't see any "assert" statements, which are what give us an actual test. Let's look into whether Mocha has a built-in way to test for the visibility of the warning div, or for an error being thrown and caught. I'm not quite sure how to get that set up but let's think about and research how we might manage it.

test/stringInputValidationTest.js
M130: Sorry, I forgot to that comment out.

jbellik:
Let me know if this is ready for another review.

jbellik (changes requested):
- Make a larger, separate div to put tests in on interface so that the width isn't restricted.
- Look into using done() to avoid timeouts.

M130:
Auto scrolling for text boxes added and also manual timeout added that didn't use done() but have put comments on how to use done() if wanted

jbellik (approved):

Issue #551, opened 2021-01-27
Minor interface tweaks


jbellik:
Under StrongStart:
- change "Foot" to "Ft" under StrongStartHsu(k/iota)
- add syllable as a category option

Under prosodic categories for GEN:
- add syllable as an option for terminals
- add foot as an option for intermediate category
- do something about default parentheses when categories are root=w, rec=Ft, terminal=syll

jbellik:
For the last bullet point: see parenthesizeTree() in treeFormatting.js
Custom bracketing would be: [] for word boundaries, () for foot boundaries, no change to syllable represenation

jbellik:
Revision to the 2nd bulletpoint above about adding syllable as a category option: You're right, syllable already seems to be present everywhere it needs to be, after all. I think I ended up taking care of that one after making this issue.

PR #555, opened 2021-02-04
Minor interface tweaks

3ddiehead:
Added ft as recursive cat and syll as terminal cat.  I only had to change the string in the categoryBrackets variable in treeFormatting, nothing with makeTableau().  The output seems to be exactly what we were looking for.

Also the issue #551 requests to add "syll" as an option in one of the strongStarts, but I think all the ones that need it have it, let me know.


treeFormatting.js
```diff
@@ -5,10 +5,10 @@ var categoryBrackets = {
 	"xp": "[]",
 	"phi": "()",
 	"x0": ["[x0 ","]"],
-	"w": ["(w ", ")"],
+	"w": ["[", "]"],
 	"clitic": ["",""],
 	"syll": ["",""],
-	"Ft": ["", ""],
+	"Ft": ["(", ")"],
```
jbellik, line 11: Let's make this by default ["(F", ")"]. I hadn't realized there were no foot brackets defined previously!

treeFormatting.js
```diff
@@ -5,10 +5,10 @@ var categoryBrackets = {
 	"xp": "[]",
 	"phi": "()",
 	"x0": ["[x0 ","]"],
-	"w": ["(w ", ")"],
+	"w": ["[", "]"],
 	"clitic": ["",""],
 	"syll": ["",""],
```
jbellik, line 10: Change to ["(s", ")"], in case we ever need to use syll as an intermediate category.

treeFormatting.js
```diff
@@ -5,10 +5,10 @@ var categoryBrackets = {
 	"xp": "[]",
 	"phi": "()",
 	"x0": ["[x0 ","]"],
-	"w": ["(w ", ")"],
+	"w": ["[", "]"],
```
jbellik, line 8: Don't change the default bracketings for w.

jbellik (changes requested):
I think I wasn't clear enough about the goal of the change in bracketing. What I am looking for is:
1. to maintain the default bracketing (where () = phi, (.w ) = w, (F ) = Ft) for most category settings in GEN, 
2. when rootCat === 'w', recursiveCat === 'Ft', and terminalCat === 'syll' on the interface, then for the interface to call makeTableau with a distinct set of brackets as an option, using the existing backend functionality.

A good way to do this would be to define a new variable interfaceCategoryBrackets that is always passed to makeTableau() as an option when the interface calls it, and to set interfaceCategoryBrackets = categoryBrackets by default, but reset it to some other new variable, let's call it wordFtBrackets, when rootCat === 'w', recursiveCat === 'Ft', and terminalCat === 'syll'.

I hope that clarifies things. Right now, with the current pull request, the problem is that ( ) have become ambiguous between referring to phi edges vs. foot edges. Try running GEN with rootCat = 'phi', recCat = 'Ft', terminalCat = 'syll' and you'll see what I mean.

jbellik (approved):

PR #560, opened 2021-02-10
Trimming bug

3ddiehead:
Added definition of Match-Phrase from Elfner 2012 and changed "removes silent nodes" to "removes silent terminals."


jbellik (changes requested):
Looks great. I'd just like to see a test file for the tree trimming functions, too :) Doesn't need to be super long.

3ddiehead:
Tests finished.  We should talk about some of the conventions to follow when we start on compiling the tests in .js files into a single file.  I've thought on some potential problems and solutions.

treeTrimming.js
```diff
@@ -104,13 +108,23 @@ function trimDeadEndNodes(node){
...
 	}else if(attribute=="func"){
-		var tree = trimDeadEndNodes(trimFunctionalTerminals(inputTree));
+		var tree = trimFunctionalTerminals(inputTree);
+	}else{
+		tree = inputTree;
 	}
+	var i = 0;
+
+	//Must call multiple times to make sure all dead-ends are removed. If run only once, a non-x0 terminal will be removed but leave another dead end in its parent.
+	while(i < 5){
```
jbellik, line 123: Why 5 times? This number seems arbitrary. Shouldn't the max be linked to the depth of the tree?

test/treeTrimmingTest.js
```diff
@@ -0,0 +1,347 @@
...
+                    ]
+                }
+            ]
+        }
+    ]
+};
+
+function treeTrimmingTest(){
+    describe('Tree Trimming Tests', function() {
+        it('Does nothing if tree is OK', function() {
```
jbellik, line 319: It would be helpful if these trees were defined inside the it() statements so that it was easy to see the code for them, or if the text of the it() statement included a call to parenthesizeTree(stree1) (or whatever the relevant tree was) so that someone looking at the test can tell easily what the input was.

jbellik: Also as it stands, these tests are all dependent on the implementation of parenthesizeTree(). Instead of comparing the bracketed strings, change them to compare JSON objects, using JSON.stringify() on the output of removeSpecifiedNodes(). That way the tests won't all break unnecessarily the next time parenthesizeTree() gets adjusted.

jbellik (commented):
Looking good! I'd like to see 3 changes, noted in the comments on the code.

treeTrimming.js
jbellik: Alternative 1: Instead of using a constant 5, use the depth of the tree to determine the number of times to run this. You would have to write a function to determine the depth of the tree, however.

Alternative 2: Instead of using a constant 5, use a conditional that checks whether the number of leaves has changed. You would need to add an option to getLeaves that makes it category-sensitive, so that it only returns things with category x0 (for example).

3ddiehead: Resolved, now runs trimDeadEndNodes() if the leaves of the tree are not all category "x0."

test/treeTrimmingTest.js
3ddiehead: Trees now show when the it statement is clicked.
JSON.stringify() is now used to check instead of parenthesizeTrees().

jbellik (approved):
Looks good, thanks for these changes. I realized while I was working on the automated testing that we have some useful functions in test/message_funcs.js that I should've had you use, but I don't want to make you redo the tests at this point.

jbellik:
The auto testing wasn't passing because we didn't have a newline at the end of whatever js file preceded treeTrimmingTest.js in the build file, so the first line with the function declaration ended up appended to the end of a commented out line!

Also in order to have the test actually execute from the command line, we need to put a call to it at the end of the .js file.

PR #561, opened 2021-02-10
Permutation Bugfix

3ddiehead:
Just needed to split the "words" string on " ". Before this, every char in the string was used for permutation.


jbellik:
Closes #554

jbellik (changes requested):
Good fix. This addresses the immediate issue. 

I believe we need to make this function more robust in a number of other ways, too. It will help to have a proper test file. I think it would be best to make it a separate file from the normal GEN test file.

Cases to include in the test file and be sure the code addresses coherently:
- stree provided, but no words --> this works fine
- stree provided, along with conflicting words --> this disregards stree and just uses words. The function should provide a warning when that happens. There also needs to be documentation in the comments of the code about the priority given to the two different inputs.
- stree not provided, words provided --> this has been fixed by the current state of the pull request

jbellik:
- Add a comment to make it clear what we're using
- Prioritize words if both arguments are provided (for the sake of consistency with GEN)
- If stree has leaves and words was also provided and the length doesn't match, then give a warning. Use displayWarning() which can handle being called either in or out of the interface.
- Check that it's still working on the interface after this.

Feel free to refactor this function to pull out the permutation generator from the tree generating part.

candidategenerator.js
```diff
@@ -370,15 +371,28 @@ function GENwithPermutation(stree, words, options){
...
-	if(!words || words.length<leaves.length){
-		words = new Array(leaves.length);
-		for(var i in leaves){
-			words[i] = leaves[i].id;
-		}
-		//console.log(words);
+  //Display warning if:
+  //    -There are no words or leaves
+  //    -There are mismatching words and leaves
+  if(((words && !words.length) || !words) && ((leaves && !leaves.length) || !leaves)){
```
jbellik, line 383: This logic can be tidied

candidategenerator.js
```diff
@@ -370,15 +371,28 @@ function GENwithPermutation(stree, words, options){
...
+    displayWarning("GENwithPermutation() was not given any syntactic tree or words to permute.");
+    words = [];
+  }
+	else if(words && words.length){
+    (leaves.length && leaves.length !== words.length) ? displayWarning("GENwithPermutation() was given a mismatching syntactic tree and words as input. The words will be used, and the syntactic tree disregarded.") : "";
```
jbellik, line 388: Avoid the ternary operator where possible, since it makes code hard to read, thereby leading to errors.

test/GENargumentTests.html
```diff
@@ -0,0 +1,119 @@
...
+                },
+                {
+                    "id": "c",
+                    "cat": "x0"
+                }
+            ]
+        }
+
+        describe('GEN with various inputs', function() {
+            it('GEN with stree only', function() {
```
jbellik, line 63: Looks to me like you didn't finish updating this test file. The actual tests don't match their descriptions, and none of them are passing at the moment. Also I don't see why we should expect a length of 24 for any of these, since you haven't specified {rootCat:'phi'} or {wrapInRecCat:true}.

jbellik (changes requested):
Doesn't look quite finished, and perhaps this review is premature. I will push a couple minor changes that tidy up the logic.

jbellik (approved):

PR #562, opened 2021-02-10
BinMinHead

jbellik:
Add a "minimal" option to binMaxHead. Create binMinHead() function to call binMaxHead with {minimal: true} and add it the interface.

The creation of this constraint was motivated by discussion with Max Kaplan about a rebracketing pattern in Ojibwe, to motivate [a [b [c]] --> (a b)(c), where we need to treat left-aligned (a b) as the head. For this reason, I set the default head-marking in binMinHead to be left-aligned.


jbellik:
Redone in a new branch due to problems in test file created by merging Ed's updates to markHeads into this branch

PR #563, opened 2021-02-13
Refactored testBuild files

3ddiehead:
testBuild.bat works great, not sure about testBuild.sh.


testBuild.sh
```diff
@@ -1,4 +1,3 @@
-cat constraints/*.js > build/constraints.js
-cat build/constraints.js *.js test/*.js > build/test.js
-cat interface_JS/*.js > build/interface.js
-cat build/test.js lib/test/*.js build/interface.js > build/spot.js
+source jsbuild.sh
+cat test/*.js > build/test.js
```
jbellik, line 2: Also need to get files out of lib/test where Mocha is -- see deleted line 4

jbellik (changes requested):
Refactor looks good, but has left out the files in lib/test, so Mocha isn't loading. If you do runStringTest() from the console on the interface, it will complain about mocha not being defined.

jbellik (approved):

PR #568, opened 2021-02-21
Multiple nodes

M130:


M130:
I can take on one more issue if this is good.

jbellik (approved):
Looks like it's working! I wish we had a little automated test for this but it's not at all obvious to me how to set that up so we'll just rely on our manual testing. Thanks for sorting this out!

PR #572, opened 2021-03-02
Js attributes to g UI

M130:


jbellik (changes requested):
Looking good! One problem is that the "else if" structure has the result of only being able to add the first attribute to the GUI. If a node has silentHead:true AND func:true, only the first will be noted in the GUI. You can see this by loading the Italian built-in system and comparing the first tree in the GUI (root = cp, silentHead) vs. in the code (root = cp, silentHead, func). I advise using independent "if" statements to accumulate some amount of text to add to the basic row html.

jbellik:
I recommend taking a look at the function addAttributeLabels() in main/treeFormatting.js to see how we are handling this for parenthesizing trees. I think the use cases are different enough that it's probably not worth trying to abstract the function to be useable in both places, but it's worth thinking about.

jbellik (approved):

PR #575, opened 2021-03-07
Fixed Issue, small fix, lots of exploration

3ddiehead:
Bug fixed by passing the showHeads option into a valueless variable used to decide whether or not to call markHeads().

markHeads now flags trees after they are marked. The function only runs if the tree hasn't been touched by it before.

BinMaxHead is already connected with the GEN: Output parameters' Tree marking options but I have added info to the BinMaxHead infobox to indicate this.

I could not recreate the bug in the issue, but there was still a bug.


jbellik (changes requested):
These look like good changes. I'd like to see some additions:
- Add tableaux to the html test file to test the interface between binMaxHead and makeTableau. These should confirm that heads show up in different places depending on the setting for which side heads are on. An example of an input that should get a violation if heads are on the right but not if heads are on the left is: {((a b*) (c d e*)*)*}
- If binMaxHead is included in the list of constraints for a tableau, then options.showHeads in makeTableau() should get set to whatever the value of binMaxHeads's head side option is.
- In the tableau header, the side that binMaxHead is marking heads on should be recorded. I.e., binMaxHead-left(phi) vs. binMaxHead-right(phi). We already do this with options for the Match constraints so I'm not sure why it's not automatically happening with other constraints that have options as well.
- I think there should actually be a way of setting up Mocha tests for how binMaxHead and makeTableau handle head marking information. I'd like to have some Mocha tests for that. They can go in binMaxHeadTest() or in a test function for makeTableau (do we have one of those?!). The thing to do is to call makeTableau() with binMaxHead as one of the constraints, and then examine the output of makeTableau() to check that heads are labeled appropriately and that the header's representation of binMaxHead make a note of which side heads are labeled on.
I know that's kind of a lot -- let's discuss on Wednesday if you have questions.

jbellik:
- When "mark prosodic heads" gets auto-checked by selection of BinMaxHead, then expand the "Tree marking options" fieldset as well.
- If BinMaxHead is selected and "mark prosodic heads" hadn't already been selected, then displayWarning("You selected BinMaxHead. By default, heads will be marked on the right.")
- add a tableau to the html test file for visual inspection

jbellik:
This is looking good. One thing I noticed though is that there aren't any tests for setting side: "left". Please add one so we have something auto-checking that we haven't broken that feature. Then I think this will be done!

jbellik (approved):

PR #576, opened 2021-03-08
Tree files script

M130:
I built the shell and bat scripts and I think it is working correctly but I can't run the shell script I built because it kept stating permission denied when I tried to run it. I tried testing it from the jsbuild.sh and it seems to be running and the files are in the correct place but I am not sure how to test it from there. 

I also accidentally created duplicate branches named JS-tree-GUI and tree-build-script. I don't have the permission to delete them so if you could delete them, that would be great.


jbellik:
Ah yes, I ran into this issue the other week when I created the mochabuild.sh script. By default, .sh scripts on Unix systems are created without execution permissions. You can do a chmod on your script to give yourself execute permissions. Here's an example page about this in case you're not familiar with chmod: https://kb.iu.edu/d/abdb

I deleted those extra branches.

interface1.html
```diff
@@ -3,11 +3,11 @@
 		<title>SPOT Interface</title>
 
 		<script src="build/spot.js"></script>
```
jbellik, line 5: Need to add a line here that loads built_in_trees.js

treeBuild.sh
```diff
@@ -0,0 +1 @@
+cat treeBuild/*.js >> build/build_in_tree.js
```
jbellik, line 1: Let's call this built_in_trees.js to be more clear that the file contains many trees.

jbellik (changes requested):
Looks good overall! 3 changes requested:
- load the tree file on the interface
- change the name of the built tree file
- change permissions on treeBuild.sh (see my comment about chmod on the "Conversations" tab of this pull request)

jbellik (approved):

PR #577, opened 2021-03-10
Auto and manual combine trees

M130:


jbellik (changes requested):
This seems to be working pretty well -- good work. I think the communication with the user needs a little fine-tuning though. 
1. I don't like the way the current set up requires the user to scroll between GEN:Inputs and the Results button to see their selection take effect. Possible solutions: 
- Put the input selection div down near the Results button. 
- Add an event listener to auto-click on Results when they make a selection in the input selection dropdown menu (like how the Built-in systems menu works)
- Go back to the pop-up idea. The options could be: a) Use inputs from visible tab, b) Use inputs from both tabs, or c) Cancel
2. What does "Clear input" / "clear trees" do? I can't tell that it's doing anything right now.

Let's discuss at today's meeting.

jbellik:
I just realized, interface1.html actually does the call to built_in() directly! Take a look at line 550 of interface1.js. It uses .value to get the value rather than the displayed string. Looks like you can put the event listener directly in the html as well -- that must be what the onchange in the <select> tag is doing.

M130:
I think the user experience is a bit better with this change. There are two options for now. One is the popup option and the second one is the dropdown. The popup option kind of works but the problem is that the program runs through the rest of the program before the user click on the button. I couldn't get the popup option done with the normal Javascript because it didn't have the popup options. So I got this one done with jQuery up to the problem I describe above. The second one, the dropdown list is working. I got it to be at the bottom of the page and it gets the result when it detects the user input. There are two getInputsForTableau functions. The default one is for the dropdown and the commented out one is for the popup. Do you want to have a popup or dropdown. And when testing for the popup, you would have to delete the comment lines in interface1.html with the jQuery comment code. You should see them when you search for jQuery. Could you check them out both and let me know which one you like? Thanks!

jbellik:
Hi Su, I tried out the dropdown and I do think it's a major improvement. Let's go with that. I don't really like the idea of introducing a dependency on jquery. I know we had some pop-ups that used window.confirm() before, although I think those have all been eliminated in favor of using displayWarning().

Proposal:
- keep the dropdown with these options
- also use displayWarning() to be sure the user's attention is drawn to that dropdown. Put the same message that's in the info for that dropdown currently in the displayed warning.

A separate follow-up is to consider some additional styling on that dropdown -- although I guess the current styling is the same as what we have on the save/load. This is low-priority and doesn't need to be addressed with this pull request.

jbellik (approved):

PR #582, opened 2021-05-04
Manual inputs interface

M130:


jbellik (changes requested):
This fixes the first problem. But I realized there is a design issue. We need a more robust way to check if there are inputs on more than one tab than using "autoOrManual==0". Right now, everything works the first time, but we run into the same problem of a silent failure if we put inputs on both tabs and get results, using the menu, and then clear the trees, make new trees, and get results again. 

That is, there is a silent failure if you follow these steps:
1. Make a tree on the Manual tab
2. Make a tree on the Automatic tab
3. Click "Get results" --> The warning and dropdown menu appear
4. Select clear all trees --> all trees disappear
5. Build a new tree (either tab) --> Now there is only a tree on one tab but the "two tabs" dropdown is still present
6. Click "Get results again" --> Result: Silent failure with error in the console.

I recommend:
- a variable that checks whether there are inputs on the manual tab - manInputsPres = true/false
- a variable that checks whether there are inputs on the automatic tab - autoInputsPres = true/false
- Add a conditional to sendToTableau(e) that prevents it from continuing to execute if getInputsForTableau() doesn't return any sTrees. This should go in line 308. Right now sendToTableau() continues to execute after getInputsForTableau() returns nothing, and that's why it's continuing with checkForLongInputs() and throwing the error below every time inputs are present on both tabs:
Uncaught TypeError: Cannot read property 'length' of undefined
    at checkForLongInputs (spot.js:6334)
    at HTMLFormElement.sendToTableau (spot.js:6379)

What is the reason for using an intermediate variable autoOrManual and setting it to 1, 2, or 3, instead of directly using the value of treeSelectOption? I think this indirection makes the code more brittle and also harder to read. If possible I'd like you to eliminate that step.

M130:
Does getAutoSTreeList() saves the deleted inputs? It seems to be saying true even after clearing the Tableau.

jbellik:
@M130 Sort of. It just returns the value of sTreeList. It's only in a wrapper function for some kind of scope reason, I think. If you didn't set up clearing the trees to reset sTreeList = undefined, then sTreeList will still have whatever value it had before. sTreeList is a backend variable; its value isn't read directly off of the interface but has to be computed by autoGenInputTree().
var sTreeList;
window.getAutoSTreeList = function(){
    return sTreeList;
}

See main/interface_JS/interface_GEN_strees.js.

M130:
@jbellik Thank you!

interface_JS/interface_tableau.js
```diff
@@ -94,75 +94,104 @@ function getInputsForTableau(){
...
-    }else if (treeSelectOption == "auto-tree"){
-        autoOrManual = 2;
-    }else if(treeSelectOption == "both-tree"){
-        autoOrManual = 3;  
-    }else if(treeSelectOption == "clear-tree"){
-        clearTableau()
-        clearAll();
+    
+    //if the dropdown menu is shown and option is option-tree
+    if (treeSelectOptions == "option-tree" && document.getElementById('treeOption').style.display == "block"){
```
jbellik, line 111: Rename "option-tree" to "default"

jbellik (changes requested):
- Clarify the logic by adding "or" statements and eliminating the first "else if" block 
- Create list of test cases in a Google Doc

M130:
https://docs.google.com/document/d/1v8n2czq0GLlhx0D5YHuNrSbuv9PUbxUU2E5DwUhhcBM/edit?usp=sharing

jbellik (approved):

PR #584, opened 2021-05-13
Weak start and strong End

M130:
Both issues


M130:
I added the trees to the it statements. Instead of using parenthesizeTree for the it function description, I just put the tree there as a string as it seems easier to read. The error message is using parenthesizeTree though and not the strings.

jbellik (approved):

